PR #4, opened 2018-05-24
updated Pipenv

goern:
and all the commits from the merge..


requirements.txt
```diff
@@ -1,6 +0,0 @@
-click
-daiquiri
-pipenv
-requests
-toml
-delegator.py
```
fridex: The requirements file is needed. One can install Kebechet from PyPI - in that case requrements.txt is used.

PR #27, opened 2018-06-28
Add a generic webhook for triggering builds

fridex:


goern:
why do we need this?

fridex:
When we release Kebechet on PyPI, package releases job finds out there is a new release and triggers openshift build for us automatically so we have kebechet always up2date - see https://github.com/thoth-station/package-releases-job/pull/30

PR #28, opened 2018-06-28
Explicitly update dependencies to their latest version

fridex:


fridex:
Will merge this as I would like to proceed with work. Feel free to raise any objections on code.

PR #32, opened 2018-06-30
Do not allow pre-releases in updates

fridex:


fridex:
With this we will need to release a proper version if we would like to get updates to projects, not only release candidates (IMHO desired behavior).

PR #35, opened 2018-07-05
Fix local setup

fridex:
Move adding rsa keys after creation of .ssh directory.


goern:
We need to set the 'approved' label manually :/

PR #50, opened 2018-07-10
[WIP] started implementing a webhook receiver

goern:
so that we can use it for issue gating to mattermost...


kebechet/webhooks.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def handle_github_open_pullrequest(pullrequest: dict) -> None:
+    """Will handle with care."""
+    _LOGGER.info(f"A Pull Request has been opened: {pullrequest['url']}")
+
+    if pullrequest['title'].startswith('Automatic update of dependency'):
+        return
+
+    notify_channel(f"[{pullrequest['user']['login']}]({pullrequest['user']['url']}) just "
```
fridex: nitpicking.. `pullrequest['user']['html_url']` will be probably better here and other places :angel:

PR #86, opened 2018-07-24
Introduce manager configuration entry

fridex:


fridex:
Merging on my own, if something will go wrong, I will fix it.

PR #89, opened 2018-07-25
Introduce version manager for automatic version releases

fridex:


kebechet/managers/version/version.py
```diff
@@ -0,0 +1,69 @@
...
+
+_LOGGER = logging.getLogger(__name__)
+_VERSION_REQUEST_ISSUE = ' release'
+_VERSION_PULL_REQUEST_NAME = 'Release of version {}'
+
+
+class VersionManager(ManagerBase):
+    """Automatic version management for Python projects."""
+
+    def _adjust_version_in_sources(self, version_identifier: str):
```
goern: 👍

kebechet/source_management.py
```diff
@@ -164,7 +164,7 @@ def open_merge_request(self, commit_msg: str, branch_name: str, body: str, label
         else:
             raise NotImplementedError
 
-        merge_request.labels = set(labels)
+        merge_request.labels = set(labels or [])
```
goern: nice! this is short for `if labels == None then labels = []` ?

fridex: yes :)

fridex: It's a python idiom

PR #91, opened 2018-07-25
Let only maintainers release new packages

fridex:


config/thoth.yaml
```diff
@@ -6,9 +6,14 @@ repositories:
     managers:
       - name: update
         configuration:
-          labels: bot
+          labels: [bot]
       - name: info
       - name: version
+        configuration:
+          maintainers:
```
goern: could we keep stuff like that in an OWNERS file within the repository itself?

fridex: Sure, we can. Is it still used by Sesheta?

goern: no, right now it is not

Issue #92, opened 2018-07-26
Extend version manager with automatic semantic versioning


fridex:
I, as a package maintainer, don't want to know what is the current version of package to bump it - e.g. by opening issue "1.0.3 release". Instead, I want to just open issue "New major release", "New minor release", "New patch release" respecting semver (optionally with "release candidate" for all of them - for example "New major release candidate"). This way I don't need to explicitly say what version I want to have and the version manager will do all the version magic for me.

fridex:
This looks like exactly what we want: https://github.com/k-bx/python-semver/blob/master/semver.py

Issue #103, opened 2018-07-26
Version manager does not work


fridex:
Version manager relies on not yet released logic in igitt - we need to wait for IGitt upstream to release a new version with all the features we require.

Opened upstream issue - https://gitlab.com/gitmate/open-source/IGitt/issues/144

fridex:
Closing this as the dev release looks suitable for us.

PR #105, opened 2018-07-26
Implement semver and calver for automated releases

fridex:
Fixes: #92


kebechet/managers/version/version.py
```diff
@@ -19,28 +19,42 @@
...
+
+
+class VersionError(Exception):
+    """An exception raised on invalid version provided or found in the repo."""
 
 
 class VersionManager(ManagerBase):
     """Automatic version management for Python projects."""
 
-    @staticmethod
```
goern: why is it not a static method anymore?

fridex: Due to version computation in `_get_new_version`

fridex:
> seems to work, lets test it... btw, any unit tests?

we need to fix #103 not to work with code from master - tests can occur after that :)

PR #119, opened 2018-07-27
Introduce Pipfile requirements.txt manager

fridex:


fridex:
The link in example has to be adjusted based on real-world usage.

Issue #95, opened 2018-07-26
Start using PyPI for releases


fridex:
Currently, we install Kebechet directly from source repo that causes sometimes failures - as containers are cached, we sometimes end up with old code in container. We should start do proper releases (now with version manager, that should be transparent) and trigger OpenShift generic trigger to build based on PyPI release.

goern:
1. ja
2. cached containers sounds like an operational issue, any further hints?!

fridex:
A workaround might be:

```yaml
noCache: true
```

as we will not pin down kebechet version and always go with latest (that is triggered on new PyPI release):

goern:
can this be closed?

goern:
ping

fridex:
Yes, let's close it as we will use "active kebechet".

PR #126, opened 2018-07-28
Start using the release on PyPI

fridex:
Fixes: #95


fridex:
@goern, not sure if we want to do it now as its still not that mature to perform releases - but can help with issue tracking.

goern:
@fridex is this still relevant?

fridex:
> @fridex is this still relevant?

Yes, but let's wait with implementation improvements. We are fine with git implementation for now.

PR #137, opened 2018-07-30
added a post pipeline job to trigger a build on openshift

goern:


goern:
It might be that the URL cracks the max_line_length of Coala.. @fridex

PR #143, opened 2018-08-29
Place related before reported changelog

fridex:


fridex:
@goern Sesheta did not close PRs opened by Kebechet on new versions, I think it is due placement of related. Could you please review this?

PR #144, opened 2018-09-03
Enable change log files on Thoth repos

fridex:


fridex:
These changelog files will be automatically managed by Kebechet on releases. No human interaction needed.

PR #171, opened 2018-10-09
Propagate issue body to pull request on version updates

fridex:


fridex:
... so we can open new releases to the future, decoratively with "Depends-On: ..."

PR #187, opened 2018-11-13
Set env variables to suppress pipenv emojis

fridex:


fridex:
Otherwise we get ugly logs - see https://github.com/thoth-station/lab/issues/46

PR #199, opened 2019-02-18
added graph-sync-scheduler to configuration file

goern:
Signed-off-by: Christoph Görn <goern@redhat.com>

### DIFF
--- a/config/thoth.yaml
+++ b/config/thoth.yaml
@@ -105,6 +105,23 @@ repositories:
           labels: [bot]
           changelog_file: true
 
+  - slug: thoth-station/graph-sync-scheduler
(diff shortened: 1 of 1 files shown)


config/thoth.yaml
```diff
@@ -105,6 +105,23 @@ repositories:
...
           changelog_file: true
 
+  - slug: thoth-station/graph-sync-scheduler
+    token: '{KEBECHET_TOKEN}'
+    managers:
+      - name: update
+        configuration:
+          labels: [bot]
+      - name: info
+      - name: version
```
fridex: No need to have version manager (we do not release it on PyPI), update (and info) manager are good enough. I will update the PR.

PR #201, opened 2019-02-18
Do not use update manager in graph-sync-scheduler

fridex:


config/thoth.yaml
```diff
@@ -112,15 +112,6 @@ repositories:
         configuration:
           labels: [bot]
       - name: info
-      - name: version
```
goern: we dont need it because its not a library?

fridex: exactly

fridex:
recheck

PR #203, opened 2019-03-11
Kill Kebechet after 2 hours if it was not able compute results

fridex:


openshift/cronJob-template.yaml
```diff
@@ -82,6 +82,12 @@ objects:
                 limits:
                   memory: "2Gi"
                   cpu: "2"
+              livenessProbe:
```
goern: can we handle that signal within the app.py and send a metric to the pushgateway because of kill by kube?! so that we can have prometheus see the kill and raise an alarm if we want to?

fridex: I think this should be already handled - if liveness probe fails, there is sent SIGTERM causing sentry to sent error and if metrics are up, they should be sent by prometheus client (!?).

Maybe it would be better to handle this the other way around (let - alterting systems do it for us) set up prometheus metrics - if there is no event for X minutes, alert. This will also handle the case we have now - python interpreter is not even started.

goern: TODO lets revisit how metrics could be emitted and what metrics should be emitted

PR #215, opened 2019-06-05
Thamos manager

KPostOffice:
# Description

This is code for a Kebechet manager that works with Thoth and its user API Thamos in order to create optimal dependency stacks for the user

### This PR includes code to:
- Run a new manager
- Automatically generate a pull request on successful execution of an advise job
- Automatically open an issue if problems were found in the dependency stack
- Issues and merge requests are such that only one can exist for the current head at a time.  The current issue or merge request will always be the latest


kebechet/managers/thoth/thoth.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python3
+# Kebechet
+# Copyright(C) 2018, 2019 Fridolin Pokorny
```
fridex: @KPostOffice your name here.

kebechet/managers/thoth/thoth.py
```diff
@@ -0,0 +1,141 @@
...
+        merge_request = self.sm.open_merge_request(commit_msg, branch_name, body, labels)
+        return merge_request
+
+    @staticmethod
+    def _advise_wrap():
+        with open("Pipfile", "r") as pipfile, open("Pipfile.lock", "a+") as piplock:
+            # TODO: take into account config values for recommendation type,
+            #       limit latest versions, no static analysis, no wait,
+            #       limit, count, debug
+            res = lib.advise(pipfile.read(), piplock.read())
```
fridex: Maybe it would worth to move this to thamos repository and provide something like `lib.advise_on_repo()` which would operate on root of the git repository so we can use the same logic everywhere. WDYT?

kebechet/managers/thoth/thoth.py
```diff
@@ -0,0 +1,141 @@
...
+            self.repo = repo
+            branch_name = self._construct_branch_name()
+            branch = self.repo.git.checkout('-B', branch_name)
+            self._cached_merge_requests = self.sm.repository.merge_requests
+            if os.path.isfile('Pipfile'):
+                res = self._advise_wrap()
+                print(res)
+                if res[1] == False:
+                    self._write_advise(res)
+                    self._open_merge_request(branch_name, ['bot'], ["Pipfile.lock"])
```
fridex: Can be label taken from the `.thoth.yaml` config file stored in user's repo so users can configure the behavoir of this bot? See this default configuration file - https://github.com/thoth-station/thamos/blob/945c095a60b7562c93cda3bd1bcf67130b5c0d11/thamos/data/defaultThoth.yaml#L36-L51

KPostOffice: Yeah, you're right.  I'll either link to the `thamos` repository to give an example or copy the example from there into the README.  This gets rid of the need to add other options to the wrapper because `advise()` already uses `.thoth.yaml` in the base repository when creating advise request

kebechet/managers/thoth/thoth.py
KPostOffice: This sounds like a good idea to me, I think it's best to just keep the same default values for everything else.

goern:
Please add some description to the PR, Zak is a very good example: https://github.com/AICoE/log-anomaly-detector/pull/131

PR #220, opened 2019-06-17
Run url

KPostOffice:
## Description
This commit contains a new option for the Kebechet command line.  Instead of pointing the cli towards a configuration file locally, you can now point it at a repository on GitHub or GitLab that has a .kebechet.yaml file in its base repository

### DIFF
--- a/kebechet/config.py
+++ b/kebechet/config.py
@@ -55,11 +55,9 @@ class _Config:
     def download_conf_from_url(self, url: str, service: str):
         params = url.split("/")
         slug = "/".join(params[3:])
-        user = params[3]
-        repo = params[4]
+        token = os.environ["GIT_ACCESS_TOKEN"]
         _LOGGER.info(f"{service} service detected")
         if service == "github":
-            token = os.environ["GITHUB_TOKEN"]
             download_uri = (
                 f"https://raw.githubusercontent.com/{slug}/master/.kebechet.yaml"
             )
@@ -68,7 +66,6 @@ class _Config:
                 download_uri, headers={"Authorization": f"token {token}"}
             )
         elif service == "gitlab":
-            token = os.environ["GITLAB_TOKEN"]
             download_uri = (
                 f'https://gitlab.com/api/v4/projects/{slug.replace("/", "%2F")}' +
                 '/repository/files/.kebechet.yaml/raw?ref=master'
--- a/kebechet/config.py
+++ b/kebechet/config.py
@@ -55,14 +55,13 @@ class _Config:
             ) from exc
 
     def download_conf_from_url(self, url: str, service: str):
-        service = Service(service=service, url=url, token=os.environ["GIT_ACCESS_TOKEN"])
+        service = Service(service=service, url=url)
         tempfile = service.download_kebechet_config()
-        _LOGGER.info(tempfile.read())
-        return service.download_kebechet_config()
+        return tempfile
 
     def run_url(self, url: str, service: str):
         temp_file = self.download_conf_from_url(url, service)
-        _LOGGER.info(f"Filename = {temp_file.name}")
+        _LOGGER.info("Filename = %s", temp_file.name)
         _LOGGER.info(temp_file.read())
         self.run(temp_file.name)
 
--- a/kebechet/config.py
+++ b/kebechet/config.py
@@ -20,12 +20,14 @@
 import logging
 import os
 import yaml
+import tempfile
 
 import urllib3
 import requests
 
 from .exception import ConfigurationError
 from .enums import ServiceType
+from .utils import services
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


kebechet/config.py
```diff
@@ -46,36 +46,80 @@ def from_file(self, config_path: str):
...
+                '/repository/files/.kebechet.yaml/raw?ref=master'
+            )
+            _LOGGER.info(f"Downloading config from: {download_uri}")
+            resp = requests.get(download_uri, headers={"Private-Token": token})
+        elif service == "pagure":
+            pass  # TODO
+        else:
+            _LOGGER.warning("Service not supported")
+
+        open("temp.yaml", "wb").write(resp.content)
```
fridex: Please use the context manager. The process will still have opened this file.

It will be a good idea to utilize [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) for this - you can manage file's lifecycle in `run_url()` method. As this will run in a container, the $PWD of the process does not need to be writable.

kebechet/config.py
```diff
@@ -46,36 +46,80 @@ def from_file(self, config_path: str):
...
+                "Failed to parse configuration file: {str(exc)}"
+            ) from exc
+
+    def download_conf_from_url(self, url: str, service: str):
+        params = url.split("/")
+        slug = "/".join(params[3:])
+        user = params[3]
+        repo = params[4]
+        _LOGGER.info(f"{service} service detected")
+        if service == "github":
```
fridex: Can this be abstracted away and, say, use [this helper function](https://github.com/thoth-station/kebechet/blob/73f9aa272f3be44a7a52b9a914337bcee4154ae8/kebechet/utils.py#L69) which will ensure we will be able to add support for other services easier and in one place?

openshift/job-kebechet-template.yaml
```diff
@@ -0,0 +1,112 @@
...
+    description: This is Kebechet Job Config, this template is meant to be used by Bots, but could also be used by humans...
+    openshift.io/display-name: Kebechet Job
+    version: 1.0.0
+    tags: poc,thoth,ai-stacks,kebechet
+    template.openshift.io/documentation-url: https://github.com/thoth-station/kebechet/
+    template.openshift.io/long-description: This is Kebechet used for automatic package releases and updates.
+    template.openshift.io/provider-display-name: Red Hat, Inc.
+  labels:
+    template: kebechet
+    app: kebechet
```
fridex: Just to follow our current labeling schema:

```yaml
app: thoth
component: kebechet
```

Here and in other templates as well.

openshift/job-kebechet-template.yaml
```diff
@@ -0,0 +1,112 @@
...
+    required: true
+    name: KEBECHET_JOB_ID
+  - description: Determines if kebechet is in verbose mode
+    displayName: Verbose mode
+    required: true
+    name: KEBECHET_VERBOSE
+
+objects:
+- apiVersion: v1
+  kind: Job
```
fridex: This does not look like a correct job template to me. This looks more like a cronjob template.

kebechet/utils.py
```diff
@@ -30,6 +30,18 @@
 
 _LOGGER = logging.getLogger(__name__)
 
+# NOTE: does not support other branches
+services = {
```
fridex: From coding design point of view, it will make more sense to create a class for this and provide methods. The overall manipulation will be much straightforward and the class with methods (all can be `classmethods`) will appear in generated docs as well.

kebechet/config.py
```diff
@@ -46,36 +48,72 @@ def from_file(self, config_path: str):
...
+        token = os.environ["GIT_ACCESS_TOKEN"]
+        _LOGGER.info(f"{service} service detected")
+        download_uri = services[service]["download_url"].format(slug=slug)
+        _LOGGER.info(f"Downloading from {download_uri}")
+        auth_value = services[service]["auth"]["value"].format(token=token)
+        auth_header = services[service]["auth"]["header"]
+        resp = requests.get(download_uri, headers={auth_header: auth_value})
+        file_ = tempfile.NamedTemporaryFile()
+        file_.write(resp.content)
+        return file_
```
fridex: This all can be moved to `Service` class so you have full control there as described in my comment above. This manipulation seems to be too "low-level", hard to test and error prone.

openshift/job-kebechet-template.yaml
```diff
@@ -0,0 +1,112 @@
...
+                  app: kebechet
+                  operatingOn: ${SERVICE_NAME}
+              volumeMounts:
+              - name: ssh-config
+                mountPath: /home/user/.ssh
+              command:
+                - 'kebechet run-url'
+              args:
+                - '${REPO_URL}'
+                - '${SERVICE_NAME}'
```
fridex: Nice approach, but this will not work with the current workflow operator. Use just environment variables to pass arguments. You can follow adviser's [approach here and provide](https://github.com/thoth-station/adviser/blob/master/app.sh) `app.sh` script which will determine which subcommand to run  - [s2i compatible approach](https://github.com/sclorg/s2i-python-container/tree/master/3.6#run-strategies).

openshift/job-kebechet-template.yaml
KPostOffice: Not sure if what you mean, I tried my best to follow the advise-job template

kebechet/services.py
```diff
@@ -0,0 +1,100 @@
...
+}
+
+
+class Service():
+    """Allows for abstraction of multiple different service with easy support of my by changing `services`."""
+
+    def __init__(self, service, url, token, branch="master"):
+        """Initialize a Service object, calls constructor functions `slug_method` and `_generate_slug`."""
+        if service not in services.keys():
+            _LOGGER.warning("Service not supported")
```
fridex: No need to warn and then raise. Just raise an exception.

kebechet/services.py
```diff
@@ -0,0 +1,100 @@
...
+            raise ValueError(f"{service} is not supported at this time")
+        self.service = services[service]["service_type"]
+        _LOGGER.info(f"{service} service detected")
```
fridex:
```python
_LOGGER.info("%s service detected", service)
```

... so strings are expanded lazily. Here and in other parts. Some linters might complain.

kebechet/services.py
```diff
@@ -0,0 +1,100 @@
...
+def _keep_slug(slug: str):
+    return slug
+
+
+def _encode_fslash(slug: str):
+    return slug.replace("/", "%2F")
+
+
+# NOTE: config_url could be generated using construct_raw_file_url but I feel this makes the code easier to follow
+services = {
```
fridex: I would make `services` (private class var `_SERVICES`), `_encode_fslash` and `_keep_slug` (private methods) part of `Service`. They semantically corresponded.

kebechet/services.py
KPostOffice: I was following conventions that I found elsewhere in the code like here:
https://github.com/thoth-station/kebechet/blob/dca3392af69fb9e7726325716966cbc7f221d90e/kebechet/config.py#L170-L172

Should all code be updated to this % style?

fridex: It would be better, but don't bother with old code. Let's make sure the newly added one is better :)

KPostOffice: Okay

kebechet/services.py
```diff
@@ -0,0 +1,101 @@
...
+import tempfile
+import logging
+import os
+
+from .enums import ServiceType
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class Service():
```
fridex:
```python3
class Service:
```

kebechet/services.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+# Kebechet
+# Copyright(C) 2018, 2019 Kevin Postlethwait
```
fridex: Just 2019.

kebechet/config.py
```diff
@@ -46,36 +48,57 @@ def from_file(self, config_path: str):
...
+                "Failed to parse configuration file: {str(exc)}"
+            ) from exc
+
+    def download_conf_from_url(self, url: str, service: str):
+        service = Service(service=service, url=url)
+        tempfile = service.download_kebechet_config()
+        return tempfile
+
+    def run_url(self, url: str, service: str):
+        temp_file = self.download_conf_from_url(url, service)
```
fridex: Clean up after everything is done:

```python
        temp_file = self.download_conf_from_url(url, service)
        try:
            _LOGGER.info("Filename = %s", temp_file.name)
            _LOGGER.info(temp_file.read())
            self.run(temp_file.name)
        finally:
            os.remove(temp_file.name)
```

BTW from design POV, it would be better to have this logic explicit in the methods rather than relying on calllee to clean stuff (that might not happen).

fridex:
@KPostOffice can this be closed to support #222?

Issue #224, opened 2019-07-18
Error: No kind "CronJob" is registered for version "batch/v1"


CermakM:
When following the deployment documentation, I get the following error after running the `ansible-playbook` command.

`Error: No kind "CronJob" is registered for version "batch/v1"`

Full ansible task output:

```
{
    "changed": true,
    "cmd": [
        "oc",
        "new-app",
        "--namespace",
        "macermak-thoth-dev",
        "--template=macermak-thoth-dev/kebechet-cronjob"
    ],
    "delta": "0:00:01.052772",
    "end": "2019-07-18 13:00:19.309656",
    "msg": "non-zero return code",
    "rc": 1,
    "start": "2019-07-18 13:00:18.256884",
    "stderr": "    error: no kind \"CronJob\" is registered for version \"batch/v1\" in scheme \"k8s.io/kubernetes/pkg/api/legacyscheme/scheme.go:29\"",
    "stderr_lines": [
        "    error: no kind \"CronJob\" is registered for version \"batch/v1\" in scheme \"k8s.io/kubernetes/pkg/api/legacyscheme/scheme.go:29\""
    ],
    "stdout": "--> Deploying template \"macermak-thoth-dev/kebechet-cronjob\" for \"macermak-thoth-dev/kebechet-cronjob\" to project macermak-thoth-dev\n\n     Kebechet CronJob\n     ---------\n     This is Kebechet CronJob Config, this template is meant to be used by Bots, but could also be used by humans...\n\n--> Creating resources ...\n--> Failed",
    "stdout_lines": [
        "--> Deploying template \"macermak-thoth-dev/kebechet-cronjob\" for \"macermak-thoth-dev/kebechet-cronjob\" to project macermak-thoth-dev",
        "",
        "     Kebechet CronJob",
        "     ---------",
        "     This is Kebechet CronJob Config, this template is meant to be used by Bots, but could also be used by humans...",
        "",
        "--> Creating resources ...",
        "--> Failed"
    ]
}
```

CermakM:
Related? https://github.com/openshift/origin/issues/21789

CermakM:
Solved by #225. Closing.

PR #225, opened 2019-07-18
CronJobs are now in apiVersion v1beta1

fridex:
Related: #224

### DIFF
--- a/openshift/cronJob-template.yaml
+++ b/openshift/cronJob-template.yaml
@@ -16,7 +16,7 @@ metadata:
 
 
 objects:
-- apiVersion: batch/v2alpha1
+- apiVersion: batch/v1beta1
   kind: CronJob
   metadata:
     name: kebechet


CermakM:
Both of these changes have to be present, doesn't work with the original `v2alpha1`

fridex:
> Doesn't work for me either.
> 
> Works when adding this change, however (from issue [gerald1248/openshift-backup#13](https://github.com/gerald1248/openshift-backup/pull/13)):
> 

Seems relevant.

> ```
>   - name: create CronJob
>     shell: oc process --filename ../openshift/cronJob-template.yaml | oc create --namespace {{ KEBECHET_INFRA_NAMESPACE }} -f -
>     ignore_errors: true
> ```

Mind the `ignore_errors` configuration - check if it really gets created.

fridex:
> Both of these changes have to be present, doesn't work with the original `v2alpha1`

Please review this PR again in that case.

PR #242, opened 2019-09-24
Add package-analyzer to monitored repos

fridex:


fridex:
@bissenbay package updates were not performed on the repo so we run old thoth-analyzer with the serialization bug.

PR #243, opened 2019-09-24
Do not monitor conu repo

fridex:


fridex:
They are not using us anymore.

PR #249, opened 2019-10-02
Fix version conflict

### LINKED ISSUE #248: Version conflict when installed with pip causes runtime failure
**Describe the bug**
There is a version conflict which produces the following error:

<details>
<summary>Log</summary>
<pre>
2019-10-02T12:39:28.0150908Z Traceback (most recent call last):
2019-10-02T12:39:28.0152136Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 583, in _build_master
2019-10-02T12:39:28.0152364Z     ws.require(__requires__)
2019-10-02T12:39:28.0152833Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 900, in require
2019-10-02T12:39:28.0153051Z     needed = self.resolve(parse_requirements(requirements))
2019-10-02T12:39:28.0153530Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 791, in resolve
2019-10-02T12:39:28.0153749Z     raise VersionConflict(dist, req).with_context(dependent_req)
2019-10-02T12:39:28.0154341Z pkg_resources.ContextualVersionConflict: (urllib3 1.22 (/opt/app-root/lib/python3.6/site-packages), Requirement.parse('urllib3>=1.24.2'), {'kubernetes'})
2019-10-02T12:39:28.0154530Z 
2019-10-02T12:39:28.0154706Z During handling of the above exception, another exception occurred:
2019-10-02T12:39:28.0155107Z 
2019-10-02T12:39:28.0155269Z Traceback (most recent call last):
2019-10-02T12:39:28.0155702Z   File "/opt/app-root/bin/kebechet", line 6, in <module>
2019-10-02T12:39:28.0155889Z     from pkg_resources import load_entry_point
2019-10-02T12:39:28.0156485Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3250, in <module>
2019-10-02T12:39:28.0161532Z     @_call_aside
2019-10-02T12:39:28.0162139Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3234, in _call_aside
2019-10-02T12:39:28.0169668Z     f(*args, **kwargs)
2019-10-02T12:39:28.0170726Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3263, in _initialize_master_working_set
2019-10-02T12:39:28.0183741Z     working_set = WorkingSet._build_master()
2019-10-02T12:39:28.0184400Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 585, in _build_master
2019-10-02T12:39:28.0184688Z     return cls._build_from_requirements(__requires__)
2019-10-02T12:39:28.0185374Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 598, in _build_from_requirements
2019-10-02T12:39:28.0185960Z     dists = ws.resolve(reqs, Environment())
2019-10-02T12:39:28.0186474Z   File "/opt/app-root/lib/python3.6/site-packages/pkg_resources/__init__.py", line 791, in resolve
2019-10-02T12:39:28.0186691Z     raise VersionConflict(dist, req).with_context(dependent_req)
2019-10-02T12:39:28.0187220Z pkg_resources.ContextualVersionConflict: (urllib3 1.22 (/opt/app-root/lib/python3.6/site-packages), Requirement.parse('urllib3>=1.24.2'), {'kubernetes'})
</pre>
</details>

**Expected behaviour**
I expect Kebechet running without any errors.

**Additional context**
This only occurs when being installed via pip and occurs probably due to inconsistent requirements.txt or incorrectly locked down version.

Side note: thamos is also missing from the requirements.txt

CermakM:
Fixes #248

NOTE: There is still a version conflict due to `igitt`, but that does not seem to affect the runtime, might wanna check that out a bit more.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 import os
+
 from setuptools import setup, find_packages
 from pathlib import Path
 
@@ -7,7 +8,23 @@ def get_install_requires():
     with open('requirements.txt', 'r') as requirements_file:
         # TODO: respect hashes in requirements.txt file
         res = requirements_file.readlines()
-        return [req.split(' ', maxsplit=1)[0] for req in res if req]
+
+        requirements = []
+        for req in res:
+            req = req.strip('\n')
+            if not req:
+                continue
+
+            if req.startswith('-e'):  # replace with PEP-508 specification
+                url = req.strip('-e ')
+                pkg = url.split('#egg=', maxsplit=1)[-1]
+                dep = pkg + " @ " + url
+            else:
+                dep = req.split(' ', maxsplit=1)[0]
+
+            requirements.append(dep)
+
+    return requirements
 
 
 def get_version():
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -406,10 +406,10 @@
         },
         "sentry-sdk": {
             "hashes": [
-                "sha256:2529ab6f93914d01bcd80b1b16c15a025902350ab19af2033aa5ff797c1600ad",
-                "sha256:ab16b539c40cc18e3dcfec99d0d3fff16337a050eaf79d4503ef2b73af9a7d9e"
+                "sha256:15e51e74b924180c98bcd636cb4634945b0a99a124d50b433c3a9dc6a582e8db",
+                "sha256:1d6a2ee908ec6d8f96c27d78bc39e203df4d586d287c233140af7d8d1aca108a"
             ],
-            "version": "==0.12.2"
+            "version": "==0.12.3"
         },
         "six": {
             "hashes": [
@@ -453,11 +453,11 @@
         },
         "thoth-common": {
             "hashes": [
-                "sha256:a00f53c4976bca923e5865b8612ae7b4236a01e74a1f8ccb6925caf319a2a309",
-                "sha256:b3a536d4f0c8dc7327549c1a42c26433ada1d46f5d46e639868f8f0163047e6e"
+                "sha256:8fe97df6e2ff7dafc3ebb1a4cf669434b5cfc1219956c6832835844af32111dc",
+                "sha256:cc002fd4a19ddb1d96410d80d4e53fcfb061118728a3b63d8e9fef9d427543af"
             ],
             "index": "pypi",
-            "version": "==0.9.10"
+            "version": "==0.9.11"
         },
         "toml": {
             "hashes": [
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,12 @@
+-e git+https://gitlab.com/gitmate/open-source/IGitt.git@master#egg=igitt
+
 click
 daiquiri
-pipenv
-toml
 delegator.py
+pipenv
 pyyaml
-IGitt==0.4.1.dev20180722120744
-# Pinned due to igitt pinning.
-gitpython==2.1.8
-requests==2.18.4
+requests>=2.22.0  # through urllib3>=1.24.2 due to CVE-2019-11324
 semver
 sentry-sdk
 thoth-common
+toml
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,5 +8,6 @@ pyyaml
 requests>=2.22.0  # through urllib3>=1.24.2 due to CVE-2019-11324
 semver
 sentry-sdk
+thamos
 thoth-common
 toml


setup.py
```diff
@@ -7,7 +8,23 @@ def get_install_requires():
...
+            if req.startswith('-e'):  # replace with PEP-508 specification
+                url = req.strip('-e ')
+                pkg = url.split('#egg=', maxsplit=1)[-1]
+                dep = pkg + " @ " + url
+            else:
+                dep = req.split(' ', maxsplit=1)[0]
+
+            requirements.append(dep)
+
+    return requirements
```
fridex: A proper solution for this is to use `dependency_links`. 

Why there is a need to use igitt from the git repo? Is IGitt depending on urllib3?

CermakM: Why do you think that `dependency_links` is the proper solution? I am not so sure about that, especially when using pip which may run into trouble or explicitly needs `--process-dependency-links` argument (for sure in earlier versions, not 100% sure about the newer ones). There's plenty of discussions around that and I suppose it boils down to the [PEP 508] https://www.python.org/dev/peps/pep-0508/ being the accepted standard for dependency specification. You might wanna check the SO as well (for example [this](https://stackoverflow.com/questions/12518499/pip-ignores-dependency-links-in-setup-py)).
That's why I chose this approach.

As far as the igitt dependency is concerned, I am merely mimicking the entry from Pipfile to preserve consistency, that's why I used the link.

fridex: > Why do you think that `dependency_links` is the proper solution? I am not so sure about that, especially when using pip which may run into trouble or explicitly needs `--process-dependency-links` argument (for sure in earlier versions, not 100% sure about the newer ones). There's plenty of discussions around that and I suppose it boils down to the [PEP 508](https://www.python.org/dev/peps/pep-0508/) being the accepted standard for dependency specification. You might wanna check the SO as well, for example [this](https://stackoverflow.com/questions/12518499/pip-ignores-dependency-links-in-setup-py).
> That's why I chose this approach.
> 
> As far as the igitt dependency is concerned, I am merely mimicking the entry from Pipfile to preserve consistency, that's why I used the link.

You declared in requirements.txt that you install package requirements from a git repo, but then, in setup py, you basically provide igitt directly by parsing the URL. Why would you do that?

CermakM: I am not sure I understand ... I just changed the format so that it is compatible with PEP 508, it still uses the repo.

fridex: It uses the repo when Pipenv is used (as the repo is stated in Pipfile/Pipfile.lock), but the requirement specified for kebechet package is igitt (what you parsed from requirements.txt file) and the latest version is picked during installation based on requirements provided. The same thing was done before - that's why I don't see why you have made this change and how it fixes the version conflict.

CermakM: So first of all, the version conflict was fixed by pinning down the `requests` to the version which uses a compatible version of urllib3

And I am still confused about what you mean as far as the igitt is concerned, I might be missing something here.
Before, we used `
IGitt==0.4.1.dev20180722120744` in the requirements.txt file which is not what Pipfile specifies (which installs it from the repo), therefore I changed it in the requirements.txt to install it from the repo as well. However, `install_requires` does not understand the `-e ...` form, it understand the `<package> @ <url>` form according to the PEP specification I mentioned, which is what the few lines in the `get_install_requires` transforms it to.

fridex: > So first of all, the version conflict was fixed by pinning down the `requests` to the version which uses a compatible version of urllib3

OK, that explains why the igitt change is not related to version conflict.

> And I am still confused about what you mean as far as the igitt is concerned, I might be missing something here.
> Before, we used ` IGitt==0.4.1.dev20180722120744` in the requirements.txt file which is not what Pipfile specifies (which installs it from the repo), therefore I changed it in the requirements.txt to install it from the repo as well. However, `install_requires` does not understand the `-e ...` form, it understands the `<package> @ <url>` form according to the PEP specification I mentioned, which is what the few lines in the `get_install_requires` transform it to.

This links back to issue https://github.com/thoth-station/kebechet/issues/103 which introduced igitt directly from master. I tried to ping igitt upstream [here](https://gitlab.com/gitmate/open-source/IGitt/issues/144) to release a new version as we rely on not yet released logic based on the issue. Having requirements pointing to a git repo is not good - we cannot manage requirements using kebechet.

As upstream is not responsive for more than year by now, it might be good idea to drop igitt and substitute it with something which has active community and development ([ogr](https://github.com/packit-service/ogr) might be a suitable option here)

BTW igitt seems to have broken versioning scheme (see the linked issue) - might be a good idea to test with "dev" releases as they are more recent as the actual releases if we can bump igitt version there for now.

CermakM: Yep, I am up for replacing igitt completely and replace it with ogr. This would fix the remaining version conflicts as well I suppose (which are in the setuptools IIRC, which igitt pins down)

CermakM: I think we should create a separate issue for that, tho, WDYT?

CermakM:
@fridex should we merge this one and leave the igitt pointing to the head ref and take care of it later on with ogr migration or should I change that to point back to the dev release (and introduce an inconsistency with the Pipfile.lock).
Since kebechet has not been managing the requirements.txt here, I would go with merging this one and taking care of igitt ASAP, WDYT?

CermakM:
Uhm, one version conflict solves the Kebechet not running at all, but the other one causes the update manager to fail:

https://github.com/CermakM/jupyter-nbrequirements/issues/15

saisankargochhayat:
@fridex Should I close this, now that IGitt is removed?

fridex:
Yes, let's close this one.

PR #260, opened 2019-10-15
Make `v` optional in tag when matching version

### LINKED ISSUE #258: Kebechet version manager should match v* tags
It is common practice to tag releases as `vX.Y` while the version is `X.Y`.
When Kebechet computes changelog between two releases, it matches is to a version... releases tagged `v*` do not match the version exactly and an initial release is assumed.

**Expected behaviour**
The changelog is computed correctly and as a tag `vX.Y` matches version `X.Y`

CermakM:
Fixes #258

Signed-off-by: Marek Cermak <macermak@redhat.com>

modified:   kebechet/managers/version/version.py
new file:   kebechet/managers/version/version_test.py


kebechet/managers/version/version_test.py
```diff
@@ -0,0 +1,83 @@
+"""Tests for version manager."""
+
+import unittest
```
fridex: In project thoth we adopted `pytest` and `flexmock` for writing and running tests. Could be this test scenario rewritten to those frameworks?

kebechet/managers/version/version_test.py
```diff
@@ -0,0 +1,83 @@
...
+        with patch.object(FakeGit, "log") as patch_log:
+            # exact match
+            _ = self.manager._compute_changelog(
+                repo=fake_repo,
+                old_version="0.1.0",
+                new_version="0.2.0",
+                version_file=False,
+            )
+            patch_log.assert_called_with(
+                f"0.1.0..HEAD", no_merges=True, format="* %s"
```
fridex: f-strings not needed in these cases (here and bellow).

CermakM: Ah, correct! Leftover, thanks :+1:

kebechet/managers/version/version_test.py
CermakM: Right, ye sure, shouldn't be a problem.

goern:
https://github.com/thoth-station/zuul-config/blob/master/zuul.d/_pipelines.yaml#L116-L152 pre-release and release pipelines are affected by this change

CermakM:
@goern how are they affected?

CermakM:
@fridex I simply wasn't able to use flexmock for that ... I spent quite some time getting various errors from it, so I stuck with the mock and migrated at least the unittest to pytest.

goern:
> @goern how are they affected?

if we start using v-tags the pipeline's regex will not trigger on tag creation, iirc

CermakM:
We don't have to change anything, the `v` is optional in the tag.

goern:
@fridex wdyt?

goern:
@saisankargochhayat @fridex shall we merge this?!

fridex:
> @saisankargochhayat @fridex shall we merge this?!

Do we want to have `v` prefix in versions? Will it help on tekton side?

goern:
Lets do it. @saisankargochhayat can you please validate that all bots and zuul and tekton use a v-prefixed tag on git and quay?

Cc: @harshad16

Issue #263, opened 2019-10-18
Regression with recent kebechet redeployment


fridex:
**Describe the bug**

The update manager fails as there is a regression (probably with the new Pipenv release).

**To Reproduce**

See cluster logs of kebechet in aicoe-prod-bots namespace.

**Expected behavior**

Updates should be running

**Additional context**

See error message produced:

```
    main()
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/resolver.py", line 119, in main
    parsed.requirements_dir, parsed.packages)
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/resolver.py", line 85, in _main
    requirements_dir=requirements_dir,
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/resolver.py", line 69, in resolve
    req_dir=requirements_dir
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/utils.py", line 726, in resolve_deps
    req_dir=req_dir,
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/utils.py", line 480, in actually_resolve_deps
    resolved_tree = resolver.resolve()
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/utils.py", line 385, in resolve
    results = self.resolver.resolve(max_rounds=environments.PIPENV_MAX_ROUNDS)
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 102, in resolve
    has_changed, best_matches = self._resolve_one_round()
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 206, in _resolve_one_round
    for dep in self._iter_dependencies(best_match):
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 301, in _iter_dependencies
    dependencies = self.repository.get_dependencies(ireq)
[...]

PR #264, opened 2019-10-18
Adjust cache location when run in the cluster due to perms

fridex:
Fixes: #263


fridex:
Not tested, just checking where the issue can be.

goern:
cant we use `XDG_*` env vars to set the location of the pip/pipenv cache?

fridex:
> cant we use `XDG_*` env vars to set the location of the pip/pipenv cache?

I think we can configure it directly with Pipenv using `PIPENV_CACHE_DIR` - we can point it to something like `/tmp`, which is probably a better solution than this one.

PR #270, opened 2019-10-21
Fix permission issues

fridex:
As the pipenv install command creates directory, we need to explictly set its
permissions after that, otherwise permissions will be set to user 1042 which
does not match OpenShift's arbitrary user present in the container image run in
the cluster.


fridex:
Hopefully the last PR, I've tried this in the cluster.

PR #294, opened 2020-01-29
Run with managers from respective repos

### LINKED ISSUE #293: Moving managers to .thoth.yaml in respective repositories.
We would like to move the managers from `config/.thoth.yaml` to `.thoth.yaml` in their respective repositories, and just have the slug and KEBECHET_TOKEN in .thoth.yaml for now.

saisankargochhayat:
Closes #293

### DIFF
--- a/app.sh
+++ b/app.sh
@@ -22,7 +22,7 @@ export GIT_SSH_COMMAND='ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChec
 
 case $KEBECHET_SUBCOMMAND in
     'run-webhook') 
-        exec pipenv run python3 kebechet-cli run-url
+        exec pipenv run python3 kebechet-cli run-webhook
         ;;
     'run-url') 
         exec pipenv run python3 kebechet-cli run-url
--- a/app.sh
+++ b/app.sh
@@ -21,6 +21,9 @@ fi
 export GIT_SSH_COMMAND='ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'
 
 case $KEBECHET_SUBCOMMAND in
+    'run-webhook') 
+        exec pipenv run python3 kebechet-cli run-url
+        ;;
     'run-url') 
         exec pipenv run python3 kebechet-cli run-url
         ;;
--- a/config/thoth_demo.yaml
+++ b/config/thoth_demo.yaml
@@ -1,5 +1,0 @@
----
-repositories:
-  - slug: saisankargochhayat/kebechet_sample
-    token: "{KEBECHET_TOKEN}"
-    service_type: github
--- a/config/thoth_demo.yaml
+++ b/config/thoth_demo.yaml
@@ -1,0 +1,6 @@
+---
+repositories:
+  - slug: saisankargochhayat/kebechet_sample
+    token: "{KEBECHET_TOKEN}"
+    service_type: github
+    service_url: https://github.com/saisankargochhayat/kebechet_sample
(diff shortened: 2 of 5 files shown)


kebechet/config.py
```diff
@@ -290,13 +289,19 @@ def run(cls, configuration_file: str) -> None:
...
         for (
-            managers,
             slug,
             service_type,
             service_url,
             token,
             tls_verify,
         ) in config.iter_entries():
+            #cls._tls_verification(service_url, slug, verify=tls_verify)          
+            tempfile = config.download_conf_from_url(service_url, service_type)
```
saisankargochhayat: should there a new function to get data from slug and service type or something?

config/thoth_demo.yaml
```diff
@@ -0,0 +1,6 @@
+---
+repositories:
+  - slug: saisankargochhayat/kebechet_sample
+    token: "{KEBECHET_TOKEN}"
+    service_type: github
+    service_url: https://github.com/saisankargochhayat/kebechet_sample
```
saisankargochhayat: Should this be removed and just the slug name, token and service type be sufficient? 
What about the other parameters? Should we attempt reading them from the repo's .thoth.yaml?

fridex: `tls_verify` should not be taken into account in Kebechet to avoid man-in-the-middle-attacks.

`service_url` is not needed - it can be constructed out of `slug` and `service_type` in this case. For internal GitLab instances, we could keep `service_url` indeed (to reach internal gitlab instance, not the public one).

kebechet/config.py
fridex: Yes, you can get it from the repo. It should be always present in the root of the git repository and named `.thoth.yaml`.

saisankargochhayat:
@fridex So I added those changes, should I remove the managers and make individual PR's to all the repos in `.thoth.yaml`?

kebechet/config.py
```diff
@@ -35,6 +35,12 @@
 class _Config:
     """Library-wide configuration."""
 
+    # This is used to generate service urls from slugs.
+    _SERVICE_URLS = {
+        "github": {"service_url": "https://github.com/{slug}"},
+        "gitlab": {"service_url": "https://github.com/{slug}"},
```
fridex: Is there needed the nested dict?

```python
    _SERVICE_URLS = {
        "github": "https://github.com/{slug}",
        "gitlab":  "https://github.com/{slug}",
    }
```

app.sh
```diff
@@ -21,6 +21,9 @@ fi
 export GIT_SSH_COMMAND='ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'
 
 case $KEBECHET_SUBCOMMAND in
+    'run-webhook') 
+        exec pipenv run python3 kebechet-cli run-url
```
fridex:
```python
        exec pipenv run python3 kebechet-cli run-webhook
```

kebechet/config.py
goern: no, I dont think so, it could be an list? ... and there is a typo in the gitlab url

fridex: These are defaults if the host of service is not explicitly provided (for non-internal/custom deployments).

saisankargochhayat: I changed the typo. This was followed in the services.py so I thought I would use the same pattern, should I change it to something else?

app.sh
saisankargochhayat: Thanks :) Changed it.

saisankargochhayat:
The coala check seems to pass for me in local.

saisankargochhayat:
Moved to https://github.com/thoth-station/kebechet/pull/297

PR #297, opened 2020-02-05
Webhook payload as argument

### LINKED ISSUE #296: Make kebechet recieve the whole payload from a webhook.
Kebechet would receive the whole webhook payload as a string, as the only argument.

saisankargochhayat:
Closes #296


saisankargochhayat:
@fridex Can you have a look?

kebechet/cli.py
```diff
@@ -88,7 +90,28 @@ def cli_run_results(origin, service, analysis_id):
...
+            try:
+                payload = json.load(f)
+            except json.decoder.JSONDecodeError:
+                _LOGGER.error("Webhook file coudn't be parsed as a json.")
+    else:
+        # If the json is passed a string.
+        try:
+            payload = json.loads(web_payload)
+        except json.decoder.JSONDecodeError:
+            _LOGGER.error("Webhook payload coudn't be parsed.")
```
fridex: Do not catch the exception here and above. Otherwise, process silently terminates with 0 return code

kebechet/payload_parser.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, payload: dict):
+        """Initialize the parameters we require from the services."""
+        self.service_type = None
+        self.url = None
+        self.event = None
+
+        # For github webhooks
+        if 'event' in payload:
+            if payload['event'] in self._IGNORED_GITHUB_EVENTS:
+                pass
```
fridex: this pass is not needed, negate the condition.

kebechet/payload_parser.py
```diff
@@ -0,0 +1,70 @@
...
+            else:
+                github_payload = payload['payload']
+                if self._GITHUB in github_payload['sender']['url'] and payload['event'] != 'integration_installation':
+                    self.service_type = 'github'
+                    self.event = payload['event']
+                    self.github_parser(github_payload)
+        # For gitlab webhooks
+        elif 'project' in payload:
+            if self._GITLAB in payload['project']['web_url']:
+                self.service_type = 'gitlab'
```
fridex: else all values will be initialized to None?

kebechet/cli.py
saisankargochhayat: Removed try catch :+1:

kebechet/payload_parser.py
saisankargochhayat: Yes, if service_type is None, func parsed_data would return None, as we don't support that service type.

kebechet/cli.py
```diff
@@ -88,7 +90,22 @@ def cli_run_results(origin, service, analysis_id):
...
+    """Run Kebechet by providing a webhook payload."""
+    payload = None
+    if os.path.isfile(web_payload):
+        with open(web_payload) as f:
+            payload = json.load(f)
+    else:
+        # If the json is passed a string.
+        payload = json.loads(web_payload)
+    if payload:
+        config.run_webhook(payload)
```
fridex: why only if payload is not null? Should the program end silently with no operation if no payload is present?

kebechet/config.py
```diff
@@ -84,15 +85,26 @@ def download_conf_from_url(url: str, service: str):
         return tempfile
 
     @classmethod
-    def run_url(cls, url: str, service: str):
+    def run_webhook(cls, payload: dict):
+        payload = PayloadParser(payload)
+        parsed_payload = payload.parsed_data()
+        if parsed_payload:
+            cls.run_url(parsed_payload['url'], parsed_payload['service_type'], True)
```
fridex: The same as above - the program was asked to do a job.

kebechet/payload_parser.py
fridex: What about rising an exception in that case? We don't know how to handle that payload.

kebechet/cli.py
saisankargochhayat: Should we raise an exception?

fridex: Yes - if there is no payload, there is nothing to do.

kebechet/config.py
saisankargochhayat: I will raise an exception if parsed_payload is None.

kebechet/payload_parser.py
saisankargochhayat: Will do that in config.py right?

fridex: I think it corresponds to payload parsing - if the payload is unknown, the initialization of payload wrapping class should raise the exception as it cannot be properly initialized.

kebechet/cli.py
saisankargochhayat: Added an exception :)

kebechet/cli.py
```diff
@@ -88,7 +90,24 @@ def cli_run_results(origin, service, analysis_id):
...
+    else:
+        _LOGGER.exception("Payload passed to webhook handler is empty.")
```
fridex: `_LOGGER.exception` logs traceback and should be run within exception handling context (in try-except block).

To raise an exception, write something like:

```python
    if not payload:
        raise SomeExceptionType("Payload passed to webhook handler is empty.")
```

kebechet/config.py
fridex: Yes, correct. No other exception needed. :+1: Maybe the `if` statement is redundant then as `parsed_payload` always returns a non-empty dict.

kebechet/payload_parser.py
```diff
@@ -0,0 +1,73 @@
...
+                if self._GITHUB in github_payload['sender']['url']:
+                    self.service_type = 'github'
+                    self.event = payload['event']
+                    self.github_parser(github_payload)
+        # For gitlab webhooks
+        elif 'project' in payload:
+            if self._GITLAB in payload['project']['web_url']:
+                self.service_type = 'gitlab'
+        else:
+            _LOGGER.exception("Payload passed is not supported.")
```
fridex: `raise` - see my comment above.

kebechet/payload_parser.py
```diff
@@ -0,0 +1,73 @@
...
+        """Initialize the parameters we require from the services."""
+        self.service_type = None
+        self.url = None
+        self.event = None
+
+        # For github webhooks
+        if 'event' in payload:
+            if payload['event'] not in self._IGNORED_GITHUB_EVENTS:
+                github_payload = payload['payload']
+                if self._GITHUB in github_payload['sender']['url']:
```
fridex: What if I create a repo named "api.github.com"? :)

https://docs.python.org/3.7/library/urllib.parse.html?highlight=urlparse#urllib.parse.urlparse

See `netloc` attribute.

The same for gitlab case.

kebechet/cli.py
saisankargochhayat: `if not payload:
        raise WebhookPayloadError("Webhook payload is empty or cannot be parsed.")
    config.run_webhook(payload)`
Something like this?

fridex: :+1:

kebechet/payload_parser.py
saisankargochhayat: Added it :smiley:

kebechet/config.py
saisankargochhayat: Let's have this for now, as in-case, the service-type doesn't get defined? Should be have a check in `parsed_data` to see if all the parameters got defined?

PR #299, opened 2020-02-06
Add thoth-station/s2i repository

fridex:


harshad16:
would the repo s2i be a pypi package?

fridex:
> would the repo s2i be a pypi package?

Yes, named thoth-s2i (release WIP). It will also provide thoth-s2i command line.

PR #302, opened 2020-02-10
Update thoth.yaml - removed non existing repo.

saisankargochhayat:
Removed non existing repo.

### DIFF
--- a/config/thoth.yaml
+++ b/config/thoth.yaml
@@ -421,13 +421,6 @@ repositories:
           labels: [bot]
       - name: info
 
-  - slug: thoth-station/graph-sync-operator
-    token: "{KEBECHET_TOKEN}"
-    managers:
-      - name: update
-        configuration:
-          labels: [bot]
-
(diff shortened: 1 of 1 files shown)


config/thoth.yaml
```diff
@@ -421,13 +421,6 @@ repositories:
           labels: [bot]
       - name: info
 
-  - slug: thoth-station/graph-sync-operator
```
fridex: Nice catch!

The repo was renamed [1] . Could you change it to:

```yaml
  - slug: thoth-station/graph-sync-scheduler
```
[1] https://github.com/thoth-station/graph-sync-scheduler

saisankargochhayat: https://github.com/thoth-station/kebechet/blob/a485cadadd4e91ce669c1dd8faf4575f786806a1/config/thoth.yaml#L201
It already exists I guess?

fridex: :+1:

PR #303, opened 2020-02-11
Webhook payload 

saisankargochhayat:
Added the parsed dict as a class variable, based on your suggestion on the merged PR.


saisankargochhayat:
Commits moved to #304

Issue #300, opened 2020-02-10
Making manager's responsive to only particular event triggers.


saisankargochhayat:
Manager's only react if the event trigger is relevant to them.

Manager | Github Event (event_type) | Gitlab Event (object_kind)
-- | -- | --
info | ‘issues’ | ‘issue’
pipfile_requirements | ‘push’ | ‘push’, ‘merge_request’
thoth_advise | ‘push’, ‘issues' | ‘push’, ‘issues', ‘merge_request’
thoth_provenance | ‘push’, ‘issues’ | ‘push’, ‘issue’, ‘merge_request’
update | ‘push’, ‘issues’ | ‘push’, ‘issue’, ‘merge_request’
version | ‘issue’ | ‘issue’

@fridex @goern

saisankargochhayat:
Are we going to extract more info from the payload like what's the issue title or so? 
Because currently the managers themselves interact with the repo and do these operations, like 'info' looks for info in issue's.

fridex:
> @fridex @goern
> Are we going to extract more info from the payload like what's the issue title or so?
> Because currently the managers themselves interact with the repo and do these operations, like 'info' looks for info in issue's.

Yes, definitely.

This is tightly bound with [1] - the logic behind managers' implementation should decide what to do based on the event. So instead of statically declare which manager should be run on which type, this information can be part of the managers logic.

[1] https://github.com/thoth-station/kebechet/pull/304/files

saisankargochhayat:
That makes sense, I will pass the event type to the manager, and it would have it as a global variable, if it acts on this kind of event or not.

saisankargochhayat:
Should be resolved with #304

PR #304, opened 2020-02-12
Manager events

saisankargochhayat:
Address #300 
Currently there is a map for manager -> event type which is checked before running an manager in the manager implementation.
Raw payload is also stored in the manager base class.
TODO - will add documentation

### DIFF
--- a/kebechet/managers/info/info.py
+++ b/kebechet/managers/info/info.py
@@ -37,9 +37,10 @@ class InfoManager(ManagerBase):
 
     def run(self) -> typing.Optional[dict]:
         """Check for info issue and close it with a report."""
-        if self.parsed_payload.get('event') not in _EVENTS_SUPPORTED:
-            _LOGGER.info("Info manager doesn't act on %r events.", self.parsed_payload.get('event'))
-            return
+        if self.parsed_payload:
+            if self.parsed_payload.get('event') not in _EVENTS_SUPPORTED:
+                _LOGGER.info("Info manager doesn't act on %r events.", self.parsed_payload.get('event'))
+                return
         issue = self.sm.get_issue(_INFO_ISSUE_NAME)
         if not issue:
             _LOGGER.info("No issue to report to, exiting")
--- a/kebechet/managers/info/info.py
+++ b/kebechet/managers/info/info.py
@@ -28,6 +28,8 @@ from .messages import INFO_REPORT
 _INFO_ISSUE_NAME = 'Kebechet info'
 
 _LOGGER = logging.getLogger(__name__)
+# Github and Gitlab events on which the manager acts upon.
+_EVENTS_SUPPORTED = ['issues', 'issue']
 
 
 class InfoManager(ManagerBase):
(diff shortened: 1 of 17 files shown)


saisankargochhayat:
@fridex Do we need anything else on this PR? :)

kebechet/managers/manager.py
```diff
@@ -60,13 +60,17 @@ def _init_igitt(service_type: ServiceType = None, service_url: str = None) -> st
...
+                 token: str = None):
         """Initialize manager instance for talking to services."""
         self.service_type = service_type or ServiceType.GITHUB
         # This needs to be called before instantiation of SourceManagement due to changes in global variables.
         self.service_url = _init_igitt(service_type, service_url)
         # Allow token expansion from env vars.
         self.slug = slug
+        if parsed_payload:
+            self.event = parsed_payload['event']
+            self.raw_payload = parsed_payload['raw_payload']
```
fridex: I think these correspondents to the `ParsedPayload` class. Why do you duplicate it here? The same for event.

saisankargochhayat: This allows the child manager classes access these variables right? How do want to pass these values to the child classes?

fridex: `parsed_payload` is a parameter, you can access its attributes:

`parsed_payload.raw_payload`

saisankargochhayat: Umm I thought, this would not break the `run` method, the one that uses cron job, so I passed this as a separate parameter and the `if` condition.
So I will just save the `parsed_payload` as a variable in the base class right?

saisankargochhayat: @fridex I made the changes, does that look good?

kebechet/managers/info/info.py
```diff
@@ -28,13 +28,18 @@
...
+# Github and Gitlab events on which the manager acts upon.
+_EVENTS_SUPPORTED = ['issues', 'issue']
 
 
 class InfoManager(ManagerBase):
     """Manager for submitting information about running Kebechet instance."""
 
     def run(self) -> typing.Optional[dict]:
         """Check for info issue and close it with a report."""
+        if self.parsed_payload.get('event') not in _EVENTS_SUPPORTED:
```
fridex: From design/semantics PoV, it feels like the payload is parameter of a manager instead of its attribute. WDYT?

It might be easier to manipulate with managers later on (e.g. create different triggers of a manager).

saisankargochhayat: Are we supporting the cron-job run also? or the webhook is the only behaviour we need to support? 
I think we can create shallow copies of the payload from the manager base class, and work on it. I think this way the function defination's stay the same to original design, like just run(self) is all the managers need to implement. But we can change to pass the payload..

saisankargochhayat: @fridex I think I need to add the condition - 
`if parsed_payload:` for now to keep supporting all the managers on cron job.

kebechet/managers/manager.py
fridex: I don't see your point in "breaking the run method" - previously the implementation did not use webhook at all - that's why there was nothing to pass to the `run` method. Now managers will be using webhooks, so they need to know webhook payload. From design point of view - webhook is not proparty/attribute of manager, but its it parameter.

However, if you want to enforce the current implementation and its the only thing that prevents us from using gitlab, I'm fine with merging this one.

PR #312, opened 2020-02-20
Fixed ogr migration errors

saisankargochhayat:
Addressing #310 #311


kebechet/source_management.py
```diff
@@ -173,7 +173,8 @@ def open_merge_request(
             merge_request = self.repository.create_pr(
                 commit_msg, body, "master", branch_name
             )
-            merge_request.add_label(labels)
+            _LOGGER.info(labels)
```
fridex: This one seems to be not that informative in the logs.

saisankargochhayat: Oh I missed that. Let me remove it.

Issue #322, opened 2020-03-02
GitDB not found 


saisankargochhayat:
With `gitpython==1.0.1` in both requirement.txt and Pipfile.
Running it in cron job mode gives the following error - 
```
`Launching subshell in virtual environment…
 . /home/sai/.local/share/virtualenvs/kebechet-mrr_QWn5/bin/activate
[sai@sai-pc kebechet]$  . /home/sai/.local/share/virtualenvs/kebechet-mrr_QWn5/bin/activate
(kebechet) [sai@sai-pc kebechet]$ ./kebechet-cli run
Traceback (most recent call last):
  File "/home/sai/.local/share/virtualenvs/kebechet-mrr_QWn5/lib/python3.6/site-packages/git/__init__.py", line 22, in _init_externals
    import gitdb
ModuleNotFoundError: No module named 'gitdb'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./kebechet-cli", line 27, in <module>
    from kebechet.exception import WebhookPayloadError
  File "/home/sai/Downloads/thoth-station/kebechet/kebechet/__init__.py", line 4, in <module>
    from .managers import InfoManager
  File "/home/sai/Downloads/thoth-station/kebechet/kebechet/managers/__init__.py", line 3, in <module>
    from .info import InfoManager
  File "/home/sai/Downloads/thoth-station/kebechet/kebechet/managers/info/__init__.py", line 3, in <module>
    from .info import InfoManager
  File "/home/sai/Downloads/thoth-station/kebechet/kebechet/managers/info/info.py", line 23, in <module>
    from kebechet.managers.manager import ManagerBase
  File "/home/sai/Downloads/thoth-station/kebechet/kebechet/managers/manager.py", line 23, in <module>
    import git
  File "/home/sai/.local/share/virtualenvs/kebechet-mrr_QWn5/lib/python3.6/site-packages/git/__init__.py", line 30, in <module>
    _init_externals()
  File "/home/sai/.local/share/virtualenvs/kebechet-mrr_QWn5/lib/python3.6/site-packages/git/__init__.py", line 24, in _init_externals
    raise ImportError("'gitdb' could not be found in your PYTHONPATH")
ImportError: 'gitdb' could not be found in your PYTHONPATH`
```
CC - @fridex

saisankargochhayat:
So gitdb was removed as a dependency. 
https://github.com/gitpython-developers/GitPython/issues/983

PR #323, opened 2020-03-02
Fix cron job run

saisankargochhayat:
Fixing #322 
I tried with GitPython 1.0.1 I think there are issues with it and it was a 2015 release. So I resorted to version `2.1.8` that was there in our requirement.txt, however we should probably move to version 3.1.0


Pipfile
```diff
@@ -9,12 +9,13 @@ pipenv = "*"
...
 toml = "*"
 "delegator.py" = "*"
-gitpython = "==1.0.1"
+gitpython = "==2.1.8"
 pyyaml = "*"
 semver = "*"
 thoth-common = "*"
 thamos = "*"
 ogr = "*"
+gitdb2 = "==3.0.1"
```
fridex: Do we need this to be pinned?

saisankargochhayat:
I have also created another pr with the latest gitpython version #324

Pipfile
saisankargochhayat: That's what they suggest as after that it breaks.

saisankargochhayat: https://github.com/gitpython-developers/GitPython/issues/983
Alternative is to use latest gitpython version which I have made another PR for, we can close this if we use the latest version.

fridex: Let's use the more recent version then.

PR #324, opened 2020-03-02
Fix latest

saisankargochhayat:
This is using the latest version of Gitpython which doesn't need gitdb dependency. 
Issue #322


saisankargochhayat:
Moved from #323

requirements.txt
```diff
@@ -4,7 +4,7 @@ pipenv
 toml
 delegator.py
 pyyaml
-gitpython==2.1.8
+gitpython==3.1.0
```
fridex: Please match the version specifier so we run the same version in the cluster as locally.

saisankargochhayat:
ERROR: python-gitlab 2.0.1 has requirement requests>=2.22.0, but you'll have requests 2.18.4 which is incompatible.
ERROR: kubernetes 10.0.1 has requirement urllib3>=1.24.2, but you'll have urllib3 1.22 which is incompatible.
Also these errros come up when I try to `pip install -r requirements.txt`

saisankargochhayat:
@fridex Should I merge this?

PR #327, opened 2020-03-05
Fixed messages and create initial lock commits

saisankargochhayat:
Fixed gitlab messages.
Fixes the commits access in `_create_initial_lock` method.

### DIFF
--- a/kebechet/managers/update/messages.py
+++ b/kebechet/managers/update/messages.py
@@ -81,6 +81,23 @@ Once this issue is resolved, the issue will be automatically closed by bot.
 
 """
 
+ISSUE_PIPENV_UPDATE_ALL_GITLAB = \
+    """
+Automatic dependency update failed for the current master with SHA {sha}.
+
+The automatic dependency management cannot continue. Please fix errors reported bellow.
+
+""" + PIPENV_REPORT + """
+
+##### Notes
+
+For more information, see [Pipfile]({service_url}/{slug}/-/blob/{sha}/Pipfile) and
+ [Pipfile.lock]({service_url}/{slug}/-/blob/{sha}/Pipfile.lock).
+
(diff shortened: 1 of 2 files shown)


saisankargochhayat:
Should I create another similar function that has looks for the specific sha pip file? Because the one we have in uses master? :)
Yes, using the util function is a really good idea :+1:

fridex:
> Should I create another similar function that has looks for the specific sha pip file? Because the one we have in uses master? :)
> Yes, using the util function is a really good idea

Is it possible to use master also in case of gitlab?

saisankargochhayat:
> > Should I create another similar function that has looks for the specific sha pip file? Because the one we have in uses master? :)
> > Yes, using the util function is a really good idea
> 
> Is it possible to use master also in case of gitlab?

Yes, we can use master or links like this - https://gitlab.com/saisankargochhayat/flake_fork/-/blob/cceda849df1969d2f0fd1cc7336a700093c2ca47/Pipfile
Which would be a good idea?

fridex:
> > > Should I create another similar function that has looks for the specific sha pip file? Because the one we have in uses master? :)
> > > Yes, using the util function is a really good idea
> > 
> > 
> > Is it possible to use master also in case of gitlab?
> 
> Yes, we can use master or links like this - https://gitlab.com/saisankargochhayat/flake_fork/-/blob/cceda849df1969d2f0fd1cc7336a700093c2ca47/Pipfile
> Which would be a good idea?

As kebechet operates on master, putting directly master can work well:

  https://gitlab.com/saisankargochhayat/flake_fork/-/blob/master/Pipfile

It might be also good idea to capture commit sha that caused issues so users know which commits caused issues in already closed issues - as done when the application stack cannot be reproduced.

saisankargochhayat:
So maybe just point to the master pipfile, and also print the commit sha in the message.

PR #336, opened 2020-03-24
Gitlab cee support

saisankargochhayat:
This allows us to support gitlab cee projects. 
`gitlabcee-oauth-token, gitlab-oauth-token, ca-bundle-crt` need to be set in kebechet-secret.

### DIFF
--- a/kebechet/cli.py
+++ b/kebechet/cli.py
@@ -91,7 +91,7 @@ def cli_run_results(origin, service, analysis_id):
 @click.option("-s", "--service", envvar="KEBECHET_SERVICE_NAME")
 def cli_run_url(url, service):
     """Run Kebechet by providing url to a git repository service."""
-    config.run_url(url, service, None, True)
+    config.run_url(url=url, service_type=service, parsed_payload=None, tls_verify=True)
 
 
 @cli.command("run-webhook")
--- a/kebechet/cli.py
+++ b/kebechet/cli.py
@@ -91,7 +91,7 @@ def cli_run_results(origin, service, analysis_id):
 @click.option("-s", "--service", envvar="KEBECHET_SERVICE_NAME")
 def cli_run_url(url, service):
     """Run Kebechet by providing url to a git repository service."""
-    config.run_url(url, service, None)
+    config.run_url(url, service, None, True)
 
 
 @cli.command("run-webhook")
--- a/kebechet/config.py
+++ b/kebechet/config.py
@@ -36,6 +36,11 @@ _LOGGER = logging.getLogger(__name__)
 class _Config:
     """Library-wide configuration."""
 
+    # Add more custom deployments to here, and then to services.py.
+    _CUSTOM_DEPLOYMENTS = {
+        "gitlab.cee.redhat.com": "gitlabcee"
+    }
+
     def __init__(self):
         self._repositories = None
 
@@ -93,6 +98,10 @@ class _Config:
 
     @classmethod
     def run_url(cls, url: str, service: str, parsed_payload: dict, tls_verify: bool):
+        for instance in cls._CUSTOM_DEPLOYMENTS.keys():
+            if instance in url:
+                service = cls._CUSTOM_DEPLOYMENTS[instance]
+
         temp_file = cls.download_conf_from_url(url, service)
         _LOGGER.debug("Filename = %s", temp_file.name)
 
--- a/kebechet/services.py
+++ b/kebechet/services.py
@@ -55,6 +55,14 @@ class Service():
             "auth": {"header": "Private-Token", "value": "{token}"},
             "slug_method": _encode_fslash.__func__,
         },
+        "gitlabcee": {
+            "token": "KEBECHET_GITLABCEE_TOKEN",
+            "service_type": ServiceType.GITLAB,
+            "config_url": "https://gitlab.cee.redhat.com/api/v4/projects/{slug}/repository/files/\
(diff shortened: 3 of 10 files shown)


saisankargochhayat:
> Looks pretty good to me. Don't we have to add it as an option [here](https://github.com/thoth-station/kebechet/blob/caf3720ef08f8bb548ac0ed61e53b1554d4207c7/kebechet/services.py#L43-L58) as well as other places?

I don't think gitlab cee projects have `.thoth.yaml` so we are supporting them using a cron-job and central config.

fridex:
> > Looks pretty good to me. Don't we have to add it as an option [here](https://github.com/thoth-station/kebechet/blob/caf3720ef08f8bb548ac0ed61e53b1554d4207c7/kebechet/services.py#L43-L58) as well as other places?
> 
> I don't think gitlab cee projects have `.thoth.yaml` so we are supporting them using a cron-job and central config.

Over time, they will need to have it if they want to use kebechet. We will drop the cronjob approach completely over time.

saisankargochhayat:
Added the gitlab cee to url/webhook support.

kebechet/services.py
```diff
@@ -55,6 +55,14 @@ def _encode_fslash(slug: str):
...
             "slug_method": _encode_fslash.__func__,
         },
+        "gitlabcee": {
+            "token": "KEBECHET_GITLABCEE_TOKEN",
+            "service_type": ServiceType.GITLAB,
+            "config_url": "https://gitlab.cee.redhat.com/api/v4/projects/{slug}/repository/files/\
+.thoth.yaml/raw?ref={branch}",
+            "auth": {"header": "Private-Token", "value": "{token}"},
+            "slug_method": _encode_fslash.__func__,
+        }
```
KPostOffice: This look good for now.  If others want to use this in the future we should have some way of adding configuration which allows them to add to this structure.

kebechet/config.py
```diff
@@ -93,6 +98,10 @@ def run_webhook(cls, payload: dict):
 
     @classmethod
     def run_url(cls, url: str, service: str, parsed_payload: dict, tls_verify: bool):
+        for instance in cls._CUSTOM_DEPLOYMENTS.keys():
+            if instance in url:
+                service = cls._CUSTOM_DEPLOYMENTS[instance]
+
```
KPostOffice: Maybe we could make `service` optional so that we can do some error checking if it is `None` after going through this for loop.

saisankargochhayat: Umm maybe this way the function demand's service type, and it it's not a valid type it error's out anyway.

KPostOffice: But if we are using a custom deployment what if the `service` isn't set?

fridex: We have gitlab that is hosted by us (a custom deployment). Using "gitlab" service type should trigger logic for contacting gitlab instance by using GitLab specific API endpoints (the public one by default) and the service url, if stated, should adjust where to find the service.

saisankargochhayat: Then it resorts to the default service passed to the function, that way we support normal `gitlab` service types, and only resort to the custom instance's in the case where the URL contains, custom deployments we support.

saisankargochhayat: > We have gitlab that is hosted by us (a custom deployment). Using "gitlab" service type should trigger logic for contacting gitlab instance by using GitLab specific API endpoints (the public one by default) and the service url, if stated, should adjust where to find the service.

That's what it is doing now, it adjusts to custom deployments in case if the instance URL matches the type's we support.

saisankargochhayat:
@fridex should we merge this?

kebechet/config.py
```diff
@@ -36,6 +36,11 @@
 class _Config:
     """Library-wide configuration."""
 
+    # Add more custom deployments to here, and then to services.py.
+    _CUSTOM_DEPLOYMENTS = {
+        "gitlab.cee.redhat.com": "gitlabcee"
```
fridex, line 41: Why is this necessary here? Can we supply this in the configuration file so we don't need to rebuild the application each time we add a deployment?

kebechet/cli.py
```diff
@@ -91,7 +91,7 @@ def cli_run_results(origin, service, analysis_id):
 @click.option("-s", "--service", envvar="KEBECHET_SERVICE_NAME")
 def cli_run_url(url, service):
     """Run Kebechet by providing url to a git repository service."""
-    config.run_url(url, service, None)
+    config.run_url(url, service, None, True)
```
fridex, line 94: It's a good practice to state what parameters are passed as `None` and `True`. It helps to read the code and reduce errors.

kebechet/config.py
fridex: Hm... Ideally, we should probably determine the host from payload - maybe that's a next step, but for later...

kebechet/cli.py
saisankargochhayat: Changed it :)

kebechet/config.py
saisankargochhayat: > Why is this necessary here? Can we supply this in the configuration file so we don't need to rebuild the application each time we add a deployment?

Most likely there are other things that need to be tinkered so I didn't put it in the configuration thus leading to rebuilding, maybe if you want, I can move it to a different file to make it cleaner as we add more custom deployments.

goern:
is this something we still want?! @fridex @harshad16

PR #381, opened 2020-05-05
Fix if automatic relocking PR exists

saisankargochhayat:
Automatic relocking fails if there a existing PR pointing to a wrong version, for example here - 
https://github.com/thoth-station/storages/pull/1750
Was pointing to jsonformatter 0.2.2 which was again removed. 
This should force push the latest update to the branch.

### DIFF
--- a/kebechet/managers/update/update.py
+++ b/kebechet/managers/update/update.py
@@ -504,7 +504,7 @@ class UpdateManager(ManagerBase):
             commits = pr.get_all_commits()
             if len(commits) != 1:
                 pr.comment("There have been done changes in the original pull request (multiple commits found), "
-                             "aborting doing changes to the modified opened pull request")
+                            "aborting doing changes to the modified opened pull request")
                 return False
             if self.sha != commits[0]:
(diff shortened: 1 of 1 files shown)


kebechet/managers/update/update.py
```diff
@@ -493,9 +497,27 @@ def _relock_all(self, exc: PipenvError, labels: list) -> None:
...
-        pr_id = self.sm.open_merge_request(commit_msg, branch_name, f"Fixes: #{issue.id}", labels)
-        _LOGGER.info(f"Issued automatic dependency re-locking in PR #{pr_id} to fix issue #{issue.id}")
+
+        existing_prs = self._get_prs(branch_name)
+        if len(existing_prs) == 1:
+            pr = list(existing_prs)[0]
+            commits = pr.get_all_commits()
+            if len(commits) != 1:
+                _LOGGER.info("There have been done changes in the original pull request (multiple commits found), "
+                             "aborting doing changes to the adjusted opened pull request")
```
fridex, line 507: Will this be valuable for the user on GitHub? (Do we want to notify the user with a comment?)

saisankargochhayat: Might be useful, I will add it.

PR #394, opened 2020-05-13
use metadata option when calling thamos advise

KPostOffice:
small change to propogate metadata. 
Depends on:
* https://github.com/thoth-station/thamos/pull/411
* https://github.com/thoth-station/common/pull/811


goern:
recheck

PR #400, opened 2020-05-18
Revert "Fix if automatic relocking PR exists"

saisankargochhayat:
Reverts thoth-station/kebechet#381

PR #401, opened 2020-05-18
Fix if automatic relocking PR exists

saisankargochhayat:
Reverts thoth-station/kebechet#400 and merges #381 
Sorry about the wrong merge.

### DIFF
--- a/kebechet/managers/update/update.py
+++ b/kebechet/managers/update/update.py
@@ -392,6 +392,11 @@ class UpdateManager(ManagerBase):
                 "No dependency management found in the repo - no Pipfile nor requirements.in nor requirements-dev.in"
             )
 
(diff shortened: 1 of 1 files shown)


goern:
/assign @saisankargochhayat 

is this still valid?

saisankargochhayat:
@fridex It seems like a corner case, I am not sure do we want it or not?

fridex:
> @fridex It seems like a corner case, I am not sure do we want it or not?

I think it might be valuable. WDYT?

saisankargochhayat:
> > @fridex It seems like a corner case, I am not sure do we want it or not?
> 
> I think it might be valuable. WDYT?

Let me rebase it, we can merge it maybe :+1:

saisankargochhayat:
Moved to #486

PR #428, opened 2020-06-11
Fix semver to use non-deprecated methods. 

saisankargochhayat:
From semver 2.10 the older methods are deprecated.

### DIFF
--- a/kebechet/managers/version/version.py
+++ b/kebechet/managers/version/version.py
@@ -41,7 +41,7 @@ _NO_MAINTAINERS_ERROR = "No release maintainers stated for this repository"
 _DIRECT_VERSION_TITLE = ' release'
 _RELEASE_TITLES = {
     "new calendar release": lambda _: datetime.utcnow().strftime("%Y.%m.%d"),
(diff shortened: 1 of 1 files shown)


saisankargochhayat:
> Nice!
> 
> I think the return value will need to be casted to `str`:
> 
> ```python
> >>> import semver
> >>> str(semver.VersionInfo.parse("1.0.0").bump_major())
> '2.0.0'
> >>> semver.VersionInfo.parse("1.0.0").bump_major()
> VersionInfo(major=2, minor=0, patch=0, prerelease=None, build=None)
> ```
> 
> in:
> 
> https://github.com/thoth-station/kebechet/blob/a643ae12da5faa6f227e3f4bb22d271490ad272c/kebechet/managers/version/version.py#L158-L165

It's casted in the lambda function, right?

saisankargochhayat:
Moved the casting to the return statement

fridex:
> It's casted in the lambda function, right?

Oh, you casted directly in lambda. Both are valid, I missed that one :+1:

Issue #446, opened 2020-06-18
Kebechet fails to parse configuration file


pacospace:
**Describe the bug**
Kebechet fails to parse configuration file

**To Reproduce**

**Expected behavior**
Kebechet not failing.

**Screenshots**
inputs
```
Name:                kebechet-job-08780a73
Namespace:           thoth-prod-tensorflow
ServiceAccount:      argo
Status:              Failed
Conditions:          
 Completed           True
Created:             Thu Jun 18 09:14:53 +0200 (10 minutes ago)
Started:             Thu Jun 18 09:14:53 +0200 (10 minutes ago)
Finished:            Thu Jun 18 09:15:03 +0200 (10 minutes ago)
Duration:            10 seconds
Parameters:          
[...]

PR #447, opened 2020-06-18
Make config parsing more safe

fridex:
Related: #446

### DIFF
--- a/kebechet/config.py
+++ b/kebechet/config.py
@@ -49,7 +49,7 @@ class _Config:
                 content = config_file.read()
 
         try:
-            self._repositories = yaml.safe_load(content).pop("repositories") or []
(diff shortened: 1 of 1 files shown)


kebechet/config.py
```diff
@@ -146,7 +146,7 @@ def run_url(cls, url: str, service: str, parsed_payload: dict, tls_verify: bool)
                 )
                 continue
             _LOGGER.info(f"Running manager %r for %r", manager_name, slug)
-            manager_configuration = manager.pop("configuration", {})
```
saisankargochhayat, line 149: Wasn't this more safer? `manager.pop("configuration") or {}` gives an error when manager isn't there?

fridex: Right, it should be `get` method. Thanks for spotting this :+1:

Issue #444, opened 2020-06-17
Automatic update didn't have new hash of package


saisankargochhayat:
Semver should have new package. 
https://github.com/thoth-station/kebechet/pull/435/files
as pointed out by harshad - https://github.com/thoth-station/slo-reporter/pull/33/files

saisankargochhayat:
Seem's like doing `pipenv install requests==2.24.0 --keep-outdated` for example doesn't work as it used to in Pipenv 2018, the hashes are no more updated and that leads to a wrong Pipfile. 
Related - https://github.com/pypa/pipenv/issues/966

saisankargochhayat:
https://github.com/thoth-station/kebechet/pull/514

fridex:
I think we can close this. It's an upstream Pipenv issue.

PR #448, opened 2020-06-18
Pin older version of Pipenv during the build

fridex:
Related: #444


fridex:
Until the upstream issue gets resolved, not to block builds.

Issue #427, opened 2020-06-10
Wrong Changelog created by VersionManager


saisankargochhayat:
For example - 
https://github.com/thoth-station/common/pull/853
I am suspecting this should only happen when old_version is not found however I tested it on thamos for example - https://gist.github.com/saisankargochhayat/3cf3c1256ecb654f4c00d7d292ce634a
The changelog should be generated correctly, which happens locally and on testing, however - 
https://github.com/thoth-station/thamos/pull/442
I am not sure what's causing the whole changelog to be populated, this happened a couple of times in thoth-station/common also 
cc - @fridex any suggestions?

fridex:
It might be something fishy in version tag detection. To debug this, it's worth to check what version is detected and how old/new version is treated. Specifically this part:

https://github.com/thoth-station/kebechet/blob/a643ae12da5faa6f227e3f4bb22d271490ad272c/kebechet/managers/version/version.py#L194-L207

If Kebechet does not find version tag in git, it generates changelog from all the git logs (I'm assuming that's what happening here).

saisankargochhayat:
I tried replicating the error locally, I am not sure where are going wrong. For example if we run the script in the issue, it find's out the thamos version and creates the right changelog. So I am confused, why does that condition fail in the cluster..

fridex:
Hm, trying it locally, it works also for me. Tags are correctly propagated:

```python
>>> from kebechet.utils import cloned_repo
>>> with cloned_repo("https://github.com", "thoth-station/adviser") as repo:
...   repo.git.tag().splitlines()
... 
['0.0.0', '0.0.1', '0.0.2', '0.1.0', '0.1.1', '0.2.0', '0.3.0', '0.4.0', '0.5.0', '0.6.0', '0.7.0', '0.7.1', '0.7.2', '0.7.3', '0.7.4', '0.8.0', '0.9.0', '0.9.1', '0.9.3', '0.9.4', '0.9.5']
```

Worth checking library versions then. Strange...

fridex:
@saisankargochhayat any progress/new observations on this one?

saisankargochhayat:
@fridex  This I tried recently the logs seem to come out to be correct - https://github.com/thoth-station/common/pull/866
Like you said before the only issue could be with that part - `is_tagged_version`, however I am not able to replicate it locally. 
This script tests on storages repo - https://gist.github.com/saisankargochhayat/3cf3c1256ecb654f4c00d7d292ce634a
[...]

PR #452, opened 2020-06-19
Limit version release log to 300

saisankargochhayat:
Closes https://github.com/thoth-station/source-management/issues/14
Related #427 
Limiting max changelog length to 300 to ensure post request size is below 100kb.

### DIFF
--- a/kebechet/managers/version/version.py
+++ b/kebechet/managers/version/version.py
@@ -38,6 +38,7 @@ _VERSION_PULL_REQUEST_NAME = 'Release of version {}'
 _NO_VERSION_FOUND_ISSUE_NAME = f"No version identifier found in sources to perform a release"
 _MULTIPLE_VERSIONS_FOUND_ISSUE_NAME = f"Multiple version identifiers found in sources to perform a new release"
 _NO_MAINTAINERS_ERROR = "No release maintainers stated for this repository"
+_BODY_TRUNCATED = "The changelog body was truncated, please check CHANGELOG.md for the complete changelog."
 _DIRECT_VERSION_TITLE = ' release'
 _RELEASE_TITLES = {
     "new calendar release": lambda _: datetime.utcnow().strftime("%Y.%m.%d"),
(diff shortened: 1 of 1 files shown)


kebechet/managers/version/version.py
```diff
@@ -206,7 +208,8 @@ def _compute_changelog(repo: Repo, old_version: str, new_version: str,
             # can be in case of the very first release.
             old_version = repo.git.rev_list("HEAD", max_parents=0)
 
-        changelog = repo.git.log(f'{old_version}..HEAD', no_merges=True, format='* %s').splitlines()
+        changelog = repo.git.log(
+            f'{old_version}..HEAD', no_merges=True, format='* %s').splitlines()[:_MAX_CHANELOG_SIZE]
```
fridex, line 212: I think it would be better to move this stripping logic directly to source management - to respective create issue/create PR methods to have this available all over the place.

Also, it would be nice to mention that the body was stripped in the body itself.

saisankargochhayat: ack

saisankargochhayat: @fridex the kebechet info seems to be more than 300 lines for example, but we display it without truncating it. So I feel like limiting it on the source-management would limit on other managers which don't involve ultrahook limitation..

fridex: OK, sounds reasonable in that case. Is this PR still valid? (it was closed)

saisankargochhayat: > OK, sounds reasonable in that case. Is this PR still valid? (it was closed)

Yes, let's merge this, if we notice a bigger PR body elsewhere, I will move it to source-management :)

saisankargochhayat: Do you want me to move this stripping logic to `_construct_pr_body` method? and let the whole changelog be committed to the file?

fridex: > and let the whole changelog be committed to the file?

That sounds like a good approach :+1:

saisankargochhayat: Thanks, I have changed it

PR #458, opened 2020-06-29
Remove graph-sync-scheduler

fridex:
... as it has been archived.

### DIFF
--- a/config/thoth.yaml
+++ b/config/thoth.yaml
@@ -208,14 +208,6 @@ repositories:
           labels: [bot]
           changelog_file: true
 
-  - slug: thoth-station/graph-sync-scheduler
-    token: "{KEBECHET_TOKEN}"
-    managers:
-      - name: update
-        configuration:
-          labels: [bot]
-      - name: info
-
   - slug: thoth-station/package-analyzer
     token: "{KEBECHET_TOKEN}"
     managers:


saisankargochhayat:
Thanks, could we also uninstall the khebhut app from the repo?

fridex:
> Thanks, could we also uninstall the khebhut app from the repo?

Yes, sounds like a good idea. Does GH automatically do that?

harshad16:
> > Thanks, could we also uninstall the khebhut app from the repo?
> 
> Yes, sounds like a good idea. Does GH automatically do that?

seems like GitHub doesn't do that, all the installed apps functions on archived repos as well.

fridex:
> > > Thanks, could we also uninstall the khebhut app from the repo?
> > 
> > 
> > Yes, sounds like a good idea. Does GH automatically do that?
> 
> seems like GitHub doesn't do that, all the installed apps functions on archived repos as well.

Do we want to have that logic on our side? I see pros and cons - if a repo is archived, we do not manage it. On the other hand, restored repos from an archive will be kept uninstalled.

harshad16:
> > > > Thanks, could we also uninstall the khebhut app from the repo?
> > > 
> > > 
> > > Yes, sounds like a good idea. Does GH automatically do that?
> > 
> > 
> > seems like GitHub doesn't do that, all the installed apps functions on archived repos as well.
> 
> Do we want to have that logic on our side? I see pros and cons - if a repo is archived, we do not manage it. On the other hand, restored repos from an archive will be kept uninstalled.

the archive repo doesn't allow pull request and issues.
however it allows to clone the project, may be it would be good to have some logic to skip archived projects.

fridex:
> > > > > Thanks, could we also uninstall the khebhut app from the repo?
> > > > 
> > > > 
> > > > Yes, sounds like a good idea. Does GH automatically do that?
> > > 
> > > 
> > > seems like GitHub doesn't do that, all the installed apps functions on archived repos as well.
> > 
> > 
> > Do we want to have that logic on our side? I see pros and cons - if a repo is archived, we do not manage it. On the other hand, restored repos from an archive will be kept uninstalled.
> 
> the archive repo doesn't allow pull request and issues.

Yes, sadly we cannot leave any "goodbye" message...

saisankargochhayat:
I guess we can add logic to our payload filter that would disable the app (set - is_active to false) 
- https://developer.github.com/webhooks/event-payloads/#repository 
So that we keep track of which repos are inactive or privatized

Doing this effort be only be worth it if we want to later run a script to run a cron-job on only active repo's to update stale repo's as we discussed.

fridex:
> I guess we can add logic to our payload filter that would disable the app (set - is_active to false)
> 
> * https://developer.github.com/webhooks/event-payloads/#repository
>   So that we keep track of which repos are inactive or privatized
> 
> Doing this effort be only be worth it if we want to later run a script to run a cron-job on only active repo's to update stale repo's as we discussed.

Let's discuss this so it will not get lost.

PR #460, opened 2020-06-29
Remove build-analyzers

fridex:
... substituted with buildlog-parser


fridex:
> should we archive this project too, as been substituted?

Archived :+1:

PR #510, opened 2020-08-27
Updated to pipenv 2020.8.13 and locked

saisankargochhayat:
Related #494

PR #515, opened 2020-08-27
Revert "Updated to pipenv 2020.8.13 and locked"

saisankargochhayat:
Reverts thoth-station/kebechet#510

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 click
 daiquiri
-pipenv==2020.8.13
+pipenv
 toml
 delegator.py
 pyyaml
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 click
 daiquiri
-pipenv==2020.8.13
+pipenv
```
saisankargochhayat, line 3: should this be locked at 2018?

fridex: Yes, I think so. You can do it in a separate PR.

saisankargochhayat: Sorry I added it here.

fridex: Even better :+1:

fridex (approved):

PR #559, opened 2020-09-21
WIP Added fedora 32 image

saisankargochhayat:
## Related Issues and Dependencies
Fixes #295 
Added fedora 32 image. 

The issue with fasttext still occurs, I tried inside the venv with pip 9.0.1 we can install fast text, but pipenv install fails.

PR #567, opened 2020-09-23
Updated glyph

saisankargochhayat:
Updated glyph


fridex (approved):

harshad16:
ah, the pytest issue would get resolved after the image is fixed right?

saisankargochhayat:
> ah, the pytest issue would get resolved after the image is fixed right?

#559 should get the pytests running

Issue #574, opened 2020-09-29
Kebechet constantly updating voluptous on adviser repo


fridex:
**Describe the bug**

See PR opened by Kebechet on adviser repo: https://github.com/thoth-station/adviser/pull/1403/

**To Reproduce**
1. Go to thoth-station/adviser
2. Check commit logs https://github.com/thoth-station/adviser/commits/master
3. See multiple updates of voluptous to 0.12.0

**Expected behavior**

The update should happen just once.

saisankargochhayat:
Fixed with https://github.com/thoth-station/adviser/pull/1417

PR #575, opened 2020-09-30
Added error message for duplicate packages

saisankargochhayat:
Fixes - #574 
Added an error message issue and also close it automatically when fixed, tested on adviser fork.

### DIFF
--- a/kebechet/managers/update/messages.py
+++ b/kebechet/managers/update/messages.py
@@ -209,3 +209,22 @@ ISSUE_UNSUPPORTED_PACKAGE = """Kebechet cannot support maintaining this applicat
   ```
   </details>
   """
+
+
+ISSUE_DUPLICATE_PACKAGES = """Kebechet cannot support maintaining this application as it contain's duplicate packages\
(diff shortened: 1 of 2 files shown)


saisankargochhayat:
@fridex I am not sure if we want this any-more?

fridex:
> @fridex I am not sure if we want this any-more? Its a corner case in which the user repeats the package.

I think duplicate entries are not possible in Pipfile (TOML):

```toml
[[source]]
url = "https://pypi.org/simple"
verify_ssl = true
name = "pypi"

[packages]
flask = "<0.6"
flask = "*"

[dev-packages]

[requires]
python_version = "3.8"
```

raises:

```python
toml.decoder.TomlDecodeError: Duplicate keys! (line 15 column 1 char 228)
```

If that's for dev-packages and packages, I think this could be handled transparently by Pipenv. So +1 on closing this. We should switch to Thoth based update anyway.

saisankargochhayat:
Thanks for confirming. :+1:

PR #597, opened 2020-11-19
Update manager non-atomic updates

saisankargochhayat:
Update manager now gives singular updates instead of atomic updates. 
Example - https://github.com/saisankargochhayat/kebechet_sample/pull/125

I haven't removed some unused functions in the manager, should I? Incase pipenv 2020 gets fixed we can revert the manager back to its original state (that is version 1.1.0)

Updated pipenv==2020.11.15


saisankargochhayat:
@fridex should we merge this?

fridex (approved):
Very nice! :1st_place_medal:

PR #600, opened 2020-11-19
Add AICOE Config

saisankargochhayat:
Add CI to enable testing on test env
Closes : #603


.aicoe-ci.yaml
```diff
@@ -0,0 +1,17 @@
+check:
+  - thoth-precommit
+  - thoth-build
+build:
+  base-image: "registry.access.redhat.com/ubi8"
```
saisankargochhayat, line 5: Is the base image necessary in docker file way to build?

.aicoe-ci.yaml
```diff
@@ -0,0 +1,17 @@
...
+  build-stratergy: Dockerfile # Allowed values: Source, Dockerfile, Containerfile (default: Source)
+  dockerfile-path: Dockerfile
+  custom_tag: latest # custom tag to be push to registry
```
saisankargochhayat, line 8: Also, do I let this be 'latest'?

saisankargochhayat:
@harshad16 could you please have a look?

.aicoe-ci.yaml
harshad16: No, it is not.please remove it.

.aicoe-ci.yaml
harshad16: we normally keep only tag version, but we can have the kebechet latest tag as well.

saisankargochhayat: So just remove custom_tag key?

saisankargochhayat: Thanks I have removed it

.aicoe-ci.yaml
saisankargochhayat: Thanks

harshad16 (approved):
lgtm , thanks :100:

PR #606, opened 2020-11-20
The release PR should close the issue

saisankargochhayat:
Related - https://github.com/thoth-station/kebechet/issues/603
In Incase sesheta doesn't close it, another run could lead to a release, best to close the issue when the release PR is merged. 

@harshad16 I think 1.1.2 tag is missing.

### DIFF
--- a/kebechet/managers/version/version.py
+++ b/kebechet/managers/version/version.py
@@ -334,7 +334,7 @@ class VersionManager(ManagerBase):
         body = cls._adjust_pr_body(issue)
         truncated_changelog = changelog[:_MAX_CHANELOG_SIZE]
         body += (
-            "Related: #"
+            "Closes: #"
             + str(issue.id)
             + "\n\n```"
             + "\n\nChangelog:\n"


fridex:
I think the initial design expected the release pipeline to close the issue. I understand the issue though, it's odd. What about checking if the issue handled does not have relevant PR already? In such case, it would also handle reopening the issue.

saisankargochhayat:
> I think the initial design expected the release pipeline to close the issue. I understand the issue though, it's odd. What about checking if the issue handled does not have relevant PR already? In such case, it would also handle reopening the issue.

We can, I think the current fetch PR method we have gets open PR's, so maybe add a new method and then comb through it to find the relevant PR, I think its easier if we close it using the pr body. WDYT?

harshad16:
The sefkhet-abwy which reacted on the auto-update PR's like dependency update, the release of. Works on approving as well as , creating tag from it and then closes the issue with relevant information. 
@saisankargochhayat  for the v1.1.2 tag , i would create it , so that it is available.
anyhow, it would be hard to debug now what happened to sefkhet-abwy while processing the release pull request.

I do feel this a good approach as well to close the issue as the release is made.

saisankargochhayat:
> The sefkhet-abwy which reacted on the auto-update PR's like dependency update, the release of. Works on approving as well as , creating tag from it and then closes the issue with relevant information.
> @saisankargochhayat for the v1.1.2 tag , i would create it , so that it is available.
> anyhow, it would be hard to debug now what happened to sefkhet-abwy while processing the release pull request.
> 
> I do feel this a good approach as well to close the issue as the release is made.

Let's merge, this then, that way its githubs responsibility to close the release issue when the PR is merged :)

saisankargochhayat:
@harshad16 @fridex Going to merge this, as we are manually closing release issues, ack?

fridex:
> @harshad16 @fridex Going to merge this, as we are manually closing release issues, ack?

ack.

PR #594, opened 2020-11-16
Added modifications to use gitpython

saisankargochhayat:
## Related Issues and Dependencies
GitPython now takes the access token of the APP rather than sesheta ssh to interact with Github. #585 
The Gitpython repo object is initialized with app access token and the sesheta github token is now just used to download `.thoth.yaml`.

PR #627, opened 2020-11-25
Fix links set 

### LINKED ISSUE #609: Pipfile/.lcok links seem broken
**Describe the bug**

**To Reproduce**
1. Go to 'https://github.com/thoth-station/user-api/issues/1157'
2. Click on 'Pipfile' link
3. See error

**Expected behavior**
links should be accessible and deliver HTTP:200

**Screenshots**
image

/kind bug

saisankargochhayat:
Closes #609

### DIFF
--- a/kebechet/managers/update/update.py
+++ b/kebechet/managers/update/update.py
@@ -89,7 +89,6 @@ class UpdateManager(ManagerBase):
     def repo(self, repo: git.Repo):
         """Set repository information and all derived information needed."""
         self._repo = repo
-        self.slug = repo.remote().url.split(":", maxsplit=1)[1][: -len(".git")]
 
     @property
     def sha(self):


kebechet/managers/update/update.py
```diff
@@ -89,7 +89,6 @@ def repo(self):
     def repo(self, repo: git.Repo):
         """Set repository information and all derived information needed."""
         self._repo = repo
-        self.slug = repo.remote().url.split(":", maxsplit=1)[1][: -len(".git")]
```
fridex, line 92: Was the logic behind `repo.remote()` changed? I would expect something like `git@github.com:thoth-station/adviser` reported.

fridex: I'm missing probably context why links suddenly stopped working.

saisankargochhayat: So if you look at this - https://github.com/thoth-station/user-api/issues/1157, the Pipfile links point to the modified url from the repo link, but with gitpython change, we use temp access token to clone urls, this needed to be removed. 
In this PR I removed it from manager.py - https://github.com/thoth-station/kebechet/pull/594/files

I am not sure why there were there in setter methods, as we get the slug when manager class is initialized.

saisankargochhayat: Repo remote now is  - https://github.com/thoth-station/kebechet/blob/c232d45cf3baa10582a4ae95ae02a85b9c7fd8fc/kebechet/utils.py#L65

fridex: OK, thanks :+1:

fridex (approved):
:1st_place_medal: :1st_place_medal: :+1:

Issue #581, opened 2020-10-26
Warn if the bot does not find older tag in the git log


fridex:
**Is your feature request related to a problem? Please describe.**

See https://github.com/opendatahub-io/opendatahub-operator/pull/72#discussion_r511777552 If the bot does not find an older tag in the git log, we generate changelog from the very first commit in the git log. This sounds like a good solution if the bot does not know from which point it should compute the changelog

**Describe the solution you'd like**

Warn about the tag not being found. By providing this warning in the PR, users know that the changelog could be computed potentially in the wrong way.

**Describe alternatives you've considered**

Keep the solution as-is.

Issue #277, opened 2019-11-08
Version matching in version manager produces wrong changelog if version is touched by user


fridex:
**Describe the bug**

If a user manually touches version string in sources, there is no version tag created in the git. That means the version manager cannot find a tag from which the changelog should be computed and thus creates a changelog from the first commit (the whole git history).

**To Reproduce**
1. Create a new release manually
2. Issue a new release using Kebechet
3. Kebechet opens a PR with CHANGELOG computed

**Expected behavior**

As Kebechet uses git tags to spot the previous release and matches it with the one found in sources, Kebechet should at least warn there were some releases issued but no tag was found in the history.

As Kebechet works now on an incremental flat git history, we could also consider extending Kebechet functionality to support pre-releases (where a final release changelog would be computed from pre-releases as well). In that case, the manipulation logic will need to be revisited.

saisankargochhayat:
Related - https://github.com/thoth-station/kebechet/issues/581

saisankargochhayat:
Closed with #628

PR #628, opened 2020-11-30
Added warning if release tag is missing.

saisankargochhayat:
Closes: #581 

Example warning PR - https://github.com/saisankargochhayat/kebechet_sample/pull/136
After fixing source-code - https://github.com/saisankargochhayat/kebechet_sample/pull/137

Should I close this also - https://github.com/thoth-station/kebechet/issues/277 ?

### DIFF
--- a/kebechet/managers/version/messages.py
+++ b/kebechet/managers/version/messages.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python3
+# Kebechet
+# Copyright(C) 2020 Sai Sankar Gochhayat
+#
+# This program is free software: you can redistribute it and / or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Body of issues and pull requests automatically opened."""
+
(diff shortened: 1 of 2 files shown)


saisankargochhayat:
@fridex PTAL :smile:

kebechet/managers/version/version.py
```diff
@@ -75,6 +76,8 @@
 _EVENTS_SUPPORTED = ["issues", "issue"]
 # Maximum number of log messages in a single release. Set due to ultrahook limits.
 _MAX_CHANELOG_SIZE = 300
+# Previous release tag present
+_PREV_RELEASE_TAG = False
```
fridex, line 80: Can this be moved to the class as an instance variable rather than using a global?

kebechet/managers/version/messages.py
```diff
@@ -0,0 +1,32 @@
...
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""Body of issues and pull requests automatically opened."""
+
+RELEASE_TAG_MISSING_WARNING = """
+---
+**WARNING NOTE**
+The release version mentioned in the source-code couldn't be found in git tags, \
+    hence the release is created from the start.
+If that is not the right behavior -
```
fridex, line 25:
```suggestion
If that is not the right behavior:
```

fridex (commented):
Nice, just few minor comments left. :+1:

kebechet/managers/version/version.py
saisankargochhayat: They are being used in one of the static methods so I chose to use, we could change `_compute_changelog` to a class method and then maybe add it as an instance variable? Is that a good idea?

fridex: Yes, I think it will be better than a global variable. If it would make sense, the static method can become an instance method.

saisankargochhayat: Changed :)

fridex (approved):

Issue #584, opened 2020-10-28
Kebechet can provide context to PR, Issues that are opened from internal trigger


pacospace:
As Kebechet user,

I would like to receive context about why Kebechet is opening PR or Issues regarding changes in my software stack.

- [x] Propagate context event trigger in a message sent from Kebechet administrator (https://github.com/thoth-station/workflow-helpers/blob/master/kebechet_administrator.py).
- [x] Define standard context event trigger messages for different cases (new release, CVE, HashMismatch, MissingPackage, MissingVersion) reference: https://github.com/kubernetes/community/blob/master/sig-list.md
- [x] Add context event to message sent to Kebechet administrator. (metadata)
- [x] Make sure latest datetime is used in Kebechet administrator queries but avoid sorting in query for performance issues.
- [x] Extend investigator kebechet run url trigger to propagate context event trigger (https://github.com/thoth-station/investigator/blob/master/thoth/investigator/kebechet_run_url_trigger/investigate_kebechet_run_url_trigger.py)
- [ ] Extend common to propagate the context event trigger (https://github.com/thoth-station/common/blob/a40d96080ae929c512132900842c5dcb4380b97a/thoth/common/openshift.py#L1586)
- [ ] Extend Kebechet run url inputs to have metadata dictionary to handle future metadata that wants to be provided (https://github.com/thoth-station/kebechet/blob/fdeca5d8edddad9d29fde17130bfed5cb36a523f/kebechet/config.py#L103) (all methods in kebechet )
- [ ] the metadata are propagated into adviser metadata
- [ ] `kebechet run results` argo workflow task in adviser workflow should retrieve the context event trigger from adviser document results and open PR or Issue with standardized context.

saisankargochhayat:
Related - https://github.com/thoth-station/investigator/pull/357

KPostOffice:
Depends on: https://github.com/thoth-station/kebechet/issues/696

PR #698, opened 2021-02-18
add justification based on metadata if possible

KPostOffice:

#584 

- [ ] Yes
- [x] No

Add justification based on metadata returned from adviser

## Description

I added bodies for PRs and issues to give users more info about the reason that Thoth created them.

### DIFF
--- a/kebechet/managers/thoth_advise/messages.py
+++ b/kebechet/managers/thoth_advise/messages.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+# Kebechet
+# Copyright(C) 2020 Kevin Postlethwait
+#
+# This program is free software: you can redistribute it and / or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


kebechet/managers/thoth_advise/messages.py
```diff
@@ -0,0 +1,86 @@
...
+
+This has caused your application stack to become unsolvable. Please loosen version constraints or change your runtime \
+environment.
+"""
+
+NEW_RELEASE_PR_BODY = """
+Version {version} of {package} was released. {package} is used in your project so Thoth automatically ran advise on \
+your application stack.
+
+This PR contains an automatic update of Pipfile.lock by Kebechet thoth-advise manager.
```
pacospace, line 85: Should we divide this based on the recommendation used?

if user requested `latest` -> make sense to update 
if user requested `performance` -> make sense if the new stack gives more performance
if user requested `security` -> ??
if user requested `stable` -> ??

wdyt?

KPostOffice: We can't guarantee that the new package is included in the update

pacospace: > > We can't guarantee that the new package is included in the update
> 
> I think we can provide pipeline units in adviser that read metadata supplied (e.g. which package was newly released) and will make sure these packages are part of the lock file (if found any). Then, the overall score will decide if the lock file is better than the one already used by the user.

I think make sense, and if the score is lower than the previous one, we should not open PR right?

fridex: +1, yes.

fridex (approved):
nice :+1:

goern:
this is just blocked by Thoth: Advise still running, feel free to merge it manually...

KPostOffice:
@goern I do not have the power

Issue #715, opened 2021-04-21
Kebechet managers support overlays


pacospace:
**Is your feature request related to a problem? Please describe.**
See: https://github.com/thoth-station/kebechet/issues/707

**Describe the solution you'd like**
Kebechet managers support overlays and provide updates to all stacks.

**Describe alternatives you've considered**

**Additional context**

goern:
@KPostOffice @pacospace please push this forward! :)

KPostOffice:
I demoed this on Monday

/close

PR #760, opened 2021-06-23
advise manager with overlays

KPostOffice:

#715 

- [x] Yes
- [ ] No

Requires:
- [ ] Add runtime environment to run-results in adviser workflow

The current implementation won't break, but if we turn lack of runtime-environment into an error then it will

Write results to the directories corresponding to overlay settings.

Depends on:
- [x] thamos "advise_here" reading requirements files from overlays directory rather than top-level

### DIFF
--- a/kebechet/managers/thoth_advise/thoth_advise.py
+++ b/kebechet/managers/thoth_advise/thoth_advise.py
@@ -18,10 +18,10 @@
 """Consume Thoth Output for Kebechet auto-dependency management."""
 
 import hashlib
-import os
 import logging
 import json
 import typing
+import yaml
 from thamos import lib
 
 import git  # noqa F401
@@ -31,7 +31,7 @@ from kebechet.exception import InternalError  # noqa F401
 from kebechet.exception import PipenvError  # noqa F401
 from kebechet.utils import cloned_repo
 from kebechet.managers.manager import ManagerBase
-from thoth.common import ThothAdviserIntegrationEnum
+from thoth.common import ThothAdviserIntegrationEnum, cwd
 from thoth.common.enums import InternalTriggerEnum
 
 from .messages import (
@@ -106,6 +106,7 @@ class ThothAdviseManager(ManagerBase):
         kebechet_metadata = metadata.get(
             "kebechet_metadata"
         )  # type: typing.Optional[dict]
+
         if kebechet_metadata is not None and kebechet_metadata.get(
             "message_justification"
         ):
@@ -118,6 +119,18 @@ class ThothAdviseManager(ManagerBase):
             )
         else:
             body = DEFAULT_PR_BODY
+
+        with open(".thoth.yaml", "r") as f:
+            thoth_config = yaml.safe_load(f)
+        overlays_dir = thoth_config.get("overlays_dir")
+
+        full_name = (
+            f"{overlays_dir}/{self.runtime_environment}"
+            if overlays_dir
+            else self.runtime_environment
+        )
+
+        body = f"# Automatic Update of {full_name} runtime-environment\n" + body
 
         # Delete branch if it didn't change Pipfile.lock
         diff = self.repo.git.diff(self.project.default_branch, files)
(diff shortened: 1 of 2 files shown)


kebechet/managers/thoth_advise/thoth_advise.py
```diff
@@ -119,6 +121,11 @@ def _open_merge_request(
         else:
             body = DEFAULT_PR_BODY
 
+        body = (
+            f"# Automatic Update of {self.runtime_environment} runtime-environment\n"
```
pacospace, line 125: should we add `overlays/{runtime_envrionment_name}`?

goern: if it is happening in an overlay that is a good information

KPostOffice: Yeah, this can be added

kebechet/managers/thoth_advise/thoth_advise.py
```diff
@@ -242,6 +284,15 @@ def run(self, labels: list, analysis_id=None):
                 branch = self.repo.git.checkout("-B", branch_name)  # noqa F841
                 self._cached_merge_requests = self.project.get_pr_list()
 
+                # run results needs to pass runtime_environment to Kebechet. This should always be present. It may
+                # actually be better to raise an error here. WDYT?
```
fridex, line 288: The current schema restricts `runtime_environments` to have at least one entry so this should be fine:

https://github.com/thoth-station/thamos/blob/f7971f5acab00dd76ee01fe44aed1be82c3fa861/thamos/config.py#L119-L122

Long-term, it might be better to operate on the config instance from thamos - it will make sure the schema is valid and any config related checks can go there. kebechet can wrap calls to try-except and handle any config related issues at one place (e.g. open an issue to repo).

KPostOffice: I think we are not on the same page. `self.runtime_environment` refers to the runtime_environment name which is passed to Kebechet. This comment refers to the fact that Kebechet should always call advise with runtime environment set and the advise workflow should always schedule kebechet-run-results with the same runtime environment.  run-results theoretically should never be run without being passed a runtime_environment name.  I just want the behavior to be well defined here and was wondering if we should just default to the first entry or if we should raise an error.

fridex: Ah, now I see - thanks for the explanation. I think being explicit here is the way to go, not to silently default. Can there be a case when the runtime environment name is changed while serving this request?

BTW using thamos' config abstraction can be still good.

KPostOffice: Yeah, the user could update the runtime environment after creating a request. I believe then we will fail when writing results. I think this is the behavior we want.

goern:
@pacospace @KPostOffice please push this forward! its such a great feature!

kebechet/managers/thoth_advise/thoth_advise.py
```diff
@@ -119,6 +120,18 @@ def _open_merge_request(
...
             body = DEFAULT_PR_BODY
 
+        with open(".thoth.yaml", "r") as f:
+            thoth_config = yaml.safe_load(f)
+        overlays_dir = thoth_config.get("overlays_dir")
+
+        full_name = (
+            f"{overlays_dir}/{self.runtime_environment}"
+            if overlays_dir
+            else self.runtime_environment
```
pacospace, line 130: how is `self.runtime_environment` passed to kebechet workflow? internal trigger that is using thoth database? 

isn't the thoth.yaml always more updated (cloned from repo when kebechet is scheduled right)? shouldn't we use the runtime_environment[0] by default if there are not overlay? which is the default behaviour usually when providing an advise?

KPostOffice: https://github.com/thoth-station/kebechet/pull/760/files/81075754a206a4de40dd1c5c73f014c1273a6c19#diff-2c3f14d5ab17b60e0aa85161f150995026a575029fefa506073b7087dd948611R254-R261

Here is where we set the default self.runtime_environment if it is not passed. If there is no overlays directory it defaults to the first entry in .thoth.yaml runtime environments

kebechet/managers/thoth_advise/thoth_advise.py
```diff
@@ -212,36 +236,54 @@ def run(self, labels: list, analysis_id=None):
...
-            with cloned_repo(self, depth=1) as repo:
+            with cloned_repo(self, self.project.default_branch, depth=1) as repo:
                 self.repo = repo
                 _LOGGER.info("Using analysis results from %s", analysis_id)
                 res = lib.get_analysis_results(analysis_id)
                 branch_name = self._construct_branch_name()
                 branch = self.repo.git.checkout("-B", branch_name)  # noqa F841
                 self._cached_merge_requests = self.project.get_pr_list()
 
+                # run results needs to pass runtime_environment to Kebechet. This should always be present. It may
```
pacospace, line 282: to be removed

pacospace (commented):
one minor comment and lgtm!! 👏🏼

KPostOffice:
@pacospace PTAL

kebechet/managers/thoth_advise/thoth_advise.py
```diff
@@ -212,36 +236,52 @@ def run(self, labels: list, analysis_id=None):
...
-                )
+                with open(".thoth.yaml", "r") as f:
+                    thoth_config = yaml.safe_load(f)
+
+                if self.runtime_environment:
+                    if self.runtime_environment not in [
+                        e["name"] for e in thoth_config["runtime_environments"]
+                    ]:
+                        # This is not a warning as it is expected when users remove and change runtime_environments
+                        _LOGGER.info("Requested runtime does not exist in target repo.")
```
fridex, line 250: This might be rare case on the user side. On the other hand, if we supply bad runtime environment name no message will be reported in monitoring systems. Might be good idea to consider this as an error:

```suggestion
                        _LOGGER.error("Requested runtime does not exist in target repo.")
```

fridex (approved):
lgtm, thanks! 💯

Issue #726, opened 2021-05-24
Open Issue when Manager Fails


KPostOffice:
**Is your feature request related to a problem? Please describe.**
I'd like Kebechet to create an issue in the repository when managers fail that way there are clear action items that I can either fix myself or report upstream to Kebechet.

**Describe the solution you'd like**
https://github.com/thoth-station/kebechet/blob/cf3c2c4103fa152e54cf628d711de393bf054213/kebechet/config.py#L179

When this logic is triggered an issue should also be opened in the repository that Kebechet is acting on.

PR #771, opened 2021-07-26
open issues on repo when managers fail

KPostOffice:

#726

- [ ] Yes
- [x] No

If managers fails then an issue along with a trace will be created on the users repository so that they can either open an upstream issue with us or fix their repository.


KPostOffice:
/hold

I tested this and the basic functionality is working, but I would like to add surrounding text 
to make it more actionable to the end-user.

harshad16 (commented):
lgtm
thank, fix the precommit issue :+1:

KPostOffice:
/unhold
See here for an example 
https://github.com/KPostOffice/jupyterlab-requirements/issues/6

KPostOffice:
@harshad16 PTAL

kebechet/managers/manager.py
```diff
@@ -31,6 +31,8 @@
 from ogr.services.base import BaseGitService
 from ogr.abstract import Issue, PullRequest
 
+from kebechet import utils
```
harshad16, line 34: this is not stated in __init__.py , how do this work ?
should this be , from kebechet.utils import get_issue_by_title .

harshad16 (commented):
one minor doubt @KPostOffice

KPostOffice:
@harshad16 PTAL

kebechet/managers/manager.py
KPostOffice: This works fine, utils should be discoverable as a module by file name alone. I left it as `import utils` so the distinction between `self.get_issue_by_title` and `utils.get_issue_by_title` was more clear.

harshad16 (approved):
thanks :100:

PR #796, opened 2021-08-24
if requirements.txt does not exist it must be created

KPostOffice:

aicoe-aiops/ocp-ci-analysis#345

- [ ] Yes
- [x] No

If requirements.txt file does not exist when running pipfile_requirements manager then ogr was throwing an exception that the file did not exist. However it is possible that the first time the user installs this manager they won't have a `requirements.txt` file so we shouldn't expect to find one in their repository.


fridex (approved):
lgtm 👍🏻

PR #798, opened 2021-08-25
pull runtime_environment directly from advise results

KPostOffice:
#795 

- [ ] Yes
- [x] No

Instead of relying on passing the runtime environment we can pull the passed runtime environment directly from the advise results returned by the user-api


fridex (approved):
lgtm 👍🏻

PR #804, opened 2021-09-01
log if issue creation is disabled

KPostOffice:

#790 

- [ ] Yes
- [x] No

Handling the case when GitHub returns an exception indicating that issues are disabled on a repository.


pacospace:
can't we get metrics on this?

KPostOffice:
> can't we get metrics on this?

I'm not sure how. Kebechet is short lived and only runs for one repo. Even if we used pushed gateway we would only be able to gather metrics for a single run.

harshad16:
@KPostOffice can you please rebase

PR #809, opened 2021-09-13
make opening issues configurable and add 'kebechet' section to config

KPostOffice:

https://github.com/thoth-station/jupyterlab-requirements/issues/416

Kebechet is opening a lot of issues on some repositories. This can be seen as spam from users. Here we can make it configurable so that it can be easily shut off. (off by default)

This also adds

```yaml
kebechet:
   ...
```
block to .thoth.yaml configuration as this config does not fit in manager specific config.

- [ ] Yes
- [x] No

### DIFF
--- a/kebechet/kebechet_runners.py
+++ b/kebechet/kebechet_runners.py
@@ -192,9 +192,15 @@ def run(
             )
             instance.run(**manager_configuration)
         except Exception as exc:  # noqa F841
-            _create_issue_from_exception(
-                manager_name=manager_name, ogr_service=ogr_service, slug=slug, exc=exc
-            )
+            if (keb_config := config.config.get("kebechet")) and keb_config.get(
+                "issue_on_exception"
+            ):
+                _create_issue_from_exception(
+                    manager_name=manager_name,
+                    ogr_service=ogr_service,
+                    slug=slug,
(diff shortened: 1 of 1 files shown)


pacospace (commented):
lgtm! Thanks for this! 💯

kebechet/kebechet_runners.py
```diff
@@ -192,9 +192,15 @@ def run(
             )
             instance.run(**manager_configuration)
         except Exception as exc:  # noqa F841
-            _create_issue_from_exception(
-                manager_name=manager_name, ogr_service=ogr_service, slug=slug, exc=exc
-            )
+            if (keb_config := config.config.get("kebechet")) and keb_config.get(
```
fridex, line 195: Not very happy to see this operator used - some linters banned it and it is not that straightforward to read.

fridex: Only this comment - otherwise changes look good to me. If you want to keep the walrus operator, feel free to merge this.

codificat:
Should this be documented as well?

Also: is it possible that there are some exceptions that deserve an issue while others would better not?  (maybe a follow-up issue for fine-tuning that? but then the config option might not be a boolean?)

PR #810, opened 2021-09-14
instead of opening issues in user repo open in thoth-station/support

KPostOffice:

Instead of: https://github.com/thoth-station/kebechet/pull/809


fridex (approved):
lgtm, makes sense 👍🏻

fridex:
BTW we should see these also in Sentry

PR #816, opened 2021-09-15
get relative path of overlay currently being updated for link

KPostOffice:

Links in issues on failed update were linking to root Pipfile/Pipfile.lock rather than the overlay which is currently being updated.


harshad16 (approved):
lgtm
thanks :100:

PR #821, opened 2021-09-17
do not raise issue for conn error

KPostOffice:

Connection errors usually occur due to services being down. For example right now, user-api is down so when the kebechet advise manager runs it raises a connection error because it is unreachable. Issues should be opened for uncaught exceptions that occur due to bugs within Kebechet not due to bugs from without.

- [ ] Yes
- [x] No


fridex (approved):
lgtm 👍🏻

PR #823, opened 2021-09-20
fix some typos

goern:
closes #822 

Signed-off-by: Christoph Görn <goern@redhat.com>

PR #826, opened 2021-09-21
Adjust template message

pacospace:
Signed-off-by: Francesco Murdaca <fmurdaca@redhat.com>

See: https://github.com/thoth-station/socrates/pull/19


pacospace:
> The current message template was merged just yesterday in #823, in order to fix the formatting of the PRs like [thoth-station/socrates#19](https://github.com/thoth-station/socrates/pull/19)
> 
> The changes are not yet in prod though, so that's why you still get bad formatting in PRs.
> 
> My understanding is that the real need here is a new release, so I am going to
> /close
> this one. Let me know if that's an error

Thanks @codificat!! I didn't see that change, then I think we can release :)

PR #835, opened 2021-09-23
only create one issue per exc type

KPostOffice:

- [ ] Yes
- [x] No

Limit exceptions raised to be one per exception, per line. This should keep thoth-station/support from being overloaded with issues


harshad16 (approved):
lgtm
thanks :100:

PR #844, opened 2021-09-29
do not initialize repo with any managers

KPostOffice:

Some repositories can't receive advises because they have no Python so we shouldn't install the advise manager by default

### DIFF
--- a/kebechet/managers/config_initializer/resources/simple.thoth.yaml
+++ b/kebechet/managers/config_initializer/resources/simple.thoth.yaml
@@ -10,7 +10,4 @@ runtime_environments:
     python_version: "3.8"
     recommendation_type: latest
 
-managers:
-  - name: thoth-advise
-    configuration:
-      labels: [bot]
+managers: []


pacospace:
> ## Related Issues and Dependencies
> Some repositories can't receive advises because they have no Python so we shouldn't install the advise manager by default

can't we check the type of files? like using si-cloc functions?

KPostOffice:
> > ## Related Issues and Dependencies
> > Some repositories can't receive advises because they have no Python so we shouldn't install the advise manager by default
> 
> can't we check the type of files? like using si-cloc functions? @KPostOffice

We could... I don't know if we should though. My other worry is that users won't know the functionality of the manager which was installed on their repo.

pacospace:
> > > ## Related Issues and Dependencies
> > > Some repositories can't receive advises because they have no Python so we shouldn't install the advise manager by default
> > 
> > 
> > can't we check the type of files? like using si-cloc functions? @KPostOffice
> 
> We could... I don't know if we should though. My other worry is that users won't know the functionality of the manager which was installed on their repo.

It seems like an interesting tech talk discussion, wdyt? :)

PR #845, opened 2021-09-29
don't require locked version for old deps

KPostOffice:

https://github.com/thoth-station/support/issues/82

- [ ] Yes
- [x] No

## Description

I added the ability for version to be None for old dependency versions because it's possible to update Pipfile without relocking


erikerlandson:
would this allow boot-strapping from no Pipfile.lock to having it created and populated by the CI?

KPostOffice:
> would this allow boot-strapping from no Pipfile.lock to having it created and populated by the CI?

This functionality exists already. When there is no `Pipfile.lock` or `requirements.txt` [this function](https://github.com/thoth-station/kebechet/blob/19be23750b4eb0bb0c0697977a48b4f608ef57c4/kebechet/managers/update/update.py#L516) creates the initial lock.

PR #881, opened 2021-10-29
reraise github server exception

KPostOffice:
## Description

Sometimes we may run into server errors when sending api requests. These may be flakey by reraising we allow the retry policy to take over.


harshad16 (commented):
lgtm

PR #908, opened 2021-11-16
if we have already cloned to dir, return repo from dir

KPostOffice:

https://github.com/thoth-station/support/issues/141

- [ ] Yes
- [x] No

If the repo has already been cloned we do not need to clone it again. This assumes that any run of Kebechet will only use this function to clone the repository that it is acting on. If this is not true then we can add a line like `os.path.join(clone_dir, project_name)`

### DIFF
--- a/kebechet/utils.py
+++ b/kebechet/utils.py
@@ -91,9 +91,12 @@ def cloned_repo(manager: "ManagerBase", branch=None, **clone_kwargs):
 
     if _CLONE_DIRECTORY is not None:
         with cwd(_CLONE_DIRECTORY):
-            repo = _clone_repo_and_set_vals(
-                repo_url, _CLONE_DIRECTORY, branch, **clone_kwargs
-            )
+            try:
+                repo = _clone_repo_and_set_vals(
+                    repo_url, _CLONE_DIRECTORY, branch, **clone_kwargs
+                )
+            except git.GitCommandError:
+                repo = git.Repo(_CLONE_DIRECTORY)
             yield repo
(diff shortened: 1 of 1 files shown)


harshad16 (approved):
lgtm
thanks :100:

kebechet/utils.py
```diff
@@ -91,9 +91,12 @@ def cloned_repo(manager: "ManagerBase", branch=None, **clone_kwargs):
...
         with cwd(_CLONE_DIRECTORY):
-            repo = _clone_repo_and_set_vals(
-                repo_url, _CLONE_DIRECTORY, branch, **clone_kwargs
-            )
+            try:
+                repo = _clone_repo_and_set_vals(
+                    repo_url, _CLONE_DIRECTORY, branch, **clone_kwargs
+                )
+            except git.GitCommandError:
+                repo = git.Repo(_CLONE_DIRECTORY)
```
fridex, line 99: I'm wondering if this would work well for cases when git raises `GitCommandError` because of other issues. Would something like this cover cases when the repo is cloned?

```python
if os.path.isdir(os.path.join(_CLONE_DIRECOTRY, ".git")):
    repo = git.Repo(_CLONE_DIRECTORY)
else:
    repo = _clone_repo_and_set_vals(repo_url , _CLONE_DIRECTORY, branch, **clone_kwargs)
```

KPostOffice: Yeah, I'll change to using this method. Thanks for the suggestion Frido

PR #909, opened 2021-11-16
check if .git exists so that we don't catch all gitcommand exceptions

KPostOffice:

#908


fridex (approved):
lgtm, thanks Kevin! 👍🏻

Issue #913, opened 2021-11-17
Kebechet fails due to the git command error


harshad16:
Kebechet version: v1.6.4
```
Traceback (most recent call last):
  File "/home/user/kebechet/kebechet_runners.py", line 210, in run
    instance.run(**manager_configuration)
  File "/home/user/kebechet/managers/version/version.py", line 448, in run
    v_info = self._adjust_version_in_sources(repo, labels, issue)
  File "/home/user/kebechet/managers/version/version.py", line 151, in _adjust_version_in_sources
    repo.git.add(file_path)
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.8/site-packages/git/cmd.py", line 638, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.8/site-packages/git/cmd.py", line 1183, in _call_process
    return self.execute(call, **exec_kwargs)
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.8/site-packages/git/cmd.py", line 983, in execute
    raise GitCommandError(redacted_command, status, stderr_value, stdout_value)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(1)
  cmdline: git add ./.venv/lib/python3.8/site-packages/wheel/__init__.py
  stderr: 'The following paths are ignored by one of your .gitignore files:
.venv
hint: Use -f if you really want to add them.
hint: Turn this message off by running
hint: "git config advice.addIgnoredFile false"'
2021-11-17 15:54:49,415   1 INFO     kebechet:245: Finished management for 'thoth-station/solver-error-classfier'
```
https://console-openshift-console.apps.smaug.na.operate-first.cloud/k8s/ns/thoth-backend-prod/pods/kebechet-job-211117154832-914b642af9c5b505-2746006291/logs?container=main

_Originally posted by @harshad16 in https://github.com/thoth-station/solver-error-classfier/issues/10#issuecomment-971721071_

harshad16:
fixed with #915

PR #915, opened 2021-11-17
Clean repo after using

KPostOffice:

https://github.com/thoth-station/kebechet/issues/913

### DIFF
--- a/kebechet/utils.py
+++ b/kebechet/utils.py
@@ -91,13 +91,13 @@ def cloned_repo(manager: "ManagerBase", branch=None, **clone_kwargs):
 
     if _CLONE_DIRECTORY is not None:
         with cwd(_CLONE_DIRECTORY):
-            if os.path.isdir(os.path.join(_CLONE_DIRECTORY, ".git")):
-                repo = git.Repo(_CLONE_DIRECTORY)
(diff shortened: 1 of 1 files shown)


kebechet/utils.py
```diff
@@ -91,13 +91,13 @@ def cloned_repo(manager: "ManagerBase", branch=None, **clone_kwargs):
 
     if _CLONE_DIRECTORY is not None:
         with cwd(_CLONE_DIRECTORY):
-            if os.path.isdir(os.path.join(_CLONE_DIRECTORY, ".git")):
-                repo = git.Repo(_CLONE_DIRECTORY)
+            if os.path.isdir(os.path.join(".", ".git")):
```
harshad16, line 94: are we no longer using the `_clone_directory` ?

KPostOffice: I came across a small issue if _CLONE_DIRECTORY refers to a local dir. Because we `cwd(_CLONE_DIRECTORY)` this is effectively the same

KPostOffice: I came across a small issue if _CLONE_DIRECTORY refers to a local dir. Because we `cwd(_CLONE_DIRECTORY)` this is effectively the same

harshad16 (approved):
lgtm
thanks :100:

PR #929, opened 2021-11-25
Fix format string reference to environment name

codificat:

Fixes https://github.com/thoth-station/support/issues/147

## Description

Following the change in #867 from `environment_name` to `env_name`, a string was pending an update on the reference.

## Additional context

The string is referenced in https://github.com/thoth-station/kebechet/blob/2d65f91577637f5de75d3dcdbf324d0c81951ab0/kebechet/managers/update/update.py#L1039


fridex (approved):
lgtm, thanks 👍🏻

PR #937, opened 2021-12-01
use env variable for kebechets name

KPostOffice:

https://github.com/thoth-station/support/issues/140

- [ ] Yes
- [x] No

instead of getting username info from ogr we can get it from the environment


fridex (approved):
lgtm 👍🏻 thanks!

PR #954, opened 2022-01-12
use analysis-id 4 branch name to avoid collisions

KPostOffice:

If adviser results create a PR multiple times for different adviser runs (i.e. internal trigger) the branch names will collide

- [ ] Yes
- [x] No

I changed the `_construct_branch_name` function to use the adviser ID as only one PR for each id should ever be created


fridex (approved):
looks good to me. 👍🏻 Thanks! 💯 

(I'm wondering if there is any limit to eventually take care of)

Issue #962, opened 2022-01-14
Deliver Container Image


codificat:
Hey, AICoE-CI!

Please build and deliver the following git tag: (new try after #960)

Tag: v1.7.0

codificat:
This is failing for a similar reason as #899:

```
[pipenv.exceptions.InstallError]:     gcc: fatal error: Killed signal terminated program cc1plus
[pipenv.exceptions.InstallError]:     compilation terminated.
[pipenv.exceptions.InstallError]:     error: command '/usr/bin/gcc' failed with exit code 1
```

I believe this is a resources issue. `pipenv` does manage to install the specified deps on a local environment

PR #900, opened 2021-11-09
patch the fasttext version for build fix

harshad16:
patch the fasttext version for build fix
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

## Related Issues and Dependencies

fixes: https://github.com/thoth-station/kebechet/issues/899
fixes: https://github.com/thoth-station/kebechet/issues/898

PR #964, opened 2022-01-14
Unpin fasttext version from Pipfile

codificat:

The addition of `fasttext` to `Pipfile` was done as an attempt to fix the problem with image building, in #900.
As seen in #962 the problem is still present, and I believe it's unrelated to the `fasttext` version

Remove the explicit inclusion of a `fasttext` version in the Pipfile, as it should not be necessary.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,7 +19,6 @@ GitPython = ">=3.1.0"
 PyYAML = "*"
 typing-extensions = "*"
 ogr = "*"
-fasttext = "==0.9.1"
 
 [dev-packages]
 pytest = "*"


harshad16:
This seems to be an issue , only when running with less resource 
with the recent update of the resource v1.6.8 has been built successfully
https://quay.io/repository/thoth-station/kebechet-job?tab=tags

codificat:
I believe we should be fine now that the builder containers have more resources (https://github.com/AICoE/aicoe-ci/pull/156).
Local / smoke tests show only a slightly higher memory consumption when compiling fasttext v0.9.2:

imatge

over v0.9.1:

imatge

/reopen

codificat:
@harshad16 do we still need a hold on this?

harshad16:
no we dont 
/unhold

harshad16 (approved):
lgtm
thanks :100:

PR #970, opened 2022-01-17
Enable TLS verification in the sample default configuration

codificat:
Subject basically says it all: it's about setting `tls_verify` to `true` in the configuration initialization that the config initializer suggests.


KPostOffice:
++
/approve

PR #1012, opened 2022-02-28
handle case where no packages section exists in Pipfile

KPostOffice:

Fixes: https://github.com/thoth-station/support/issues/162

- [ ] Yes
- [x] No

Use `.get("package", {})` so that there is a good default if the section doesn't exist. This is already used below for `dev-packages`


harshad16 (approved):
lgtm
thanks :100:

PR #1013, opened 2022-03-01
place limit on chars from pipenv command. leave ~5000 for the rest of…

KPostOffice:

fixes: https://github.com/thoth-station/support/issues/186

- [ ] Yes
- [x] No

this will truncate exception output to respect (roughly) github's issue body and comment length


kebechet/exception.py
```diff
@@ -39,6 +39,30 @@ def __init__(self, command: Command, *args, **kwargs):
...
+            if curr[0] < individual_char_limit:
+                to_ret[curr[1]] = curr[2]
+
+                # distribute remaining limit, the max is to avoid divide by 0 on the final iteration
+                individual_char_limit += int(
+                    (individual_char_limit - curr[0]) / max(1, len(attrs) - i - 1)
+                )
+            else:
+                to_ret[curr[1]] = curr[2][: individual_char_limit - 3] + "..."
+
```
harshad16, line 63: thanks for adding this feature.
trying to make sense of this 

the loop would iterate for (command, std_out, std_err), can there be a case where command or std_err would take up all the `individual_char_limit`? if so then it seems one of the following options would take the most of the bits, we may have to change this a bit to modulate that.

harshad16 (changes requested):
good pr. just one clarification needed

kebechet/exception.py
KPostOffice: `individual_char_limit` is 1/3 of the total limit that way all will be able to use at least 1/3 and increases if the attributes do not use their allotted limit.

harshad16: oh right, I missed it was getting divided by the length.

harshad16 (approved):
lgtm
thanks :100:

PR #1038, opened 2022-04-01
Bugfix: bump black version and fixing git repo typing bug

Gregory-Pereira:

Related to [thoth-station/s2i-generic-data-science-notebook issue 48](https://github.com/thoth-station/s2i-generic-data-science-notebook/issues/48), although this will not address that issue. This bugfix came up in the process of addressing that issue.

- [ ] Yes
- [x] No

The package `_unicodefun` was removed from `click`, and this did not work with Black v 22.0.1. Therefore I bumped the version of Black to the next version for which it would work (v22.3.0).

Additionally typing the `self.__repo` as `git.Repo` with no default value to address tests.

### DIFF
--- a/kebechet/managers/manager.py
+++ b/kebechet/managers/manager.py
@@ -63,7 +63,7 @@ class ManagerBase:
 
         self.service = service
         self.project = service.get_project(namespace=self.owner, repo=self.repo_name)
-        self._repo = None
+        self._repo: git.Repo = None
         self.metadata = metadata
         self.runtime_environment = runtime_environment
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,7 +40,7 @@ repos:
         args: [--ignore-missing-imports]
 
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
       - id: black
 


config/thoth.yaml
```diff
@@ -664,3 +664,21 @@ repositories:
...
+            - sesheta
+            - harshad16
+            - goern
+          assignees:
+            - sesheta
+          labels: [bot]
+          changelog_file: true
+      - name: update
+        configuration:
+          labels: [bot]
```
harshad16, line 684: TIL I remember, this part was used in old days when kebechet used to run as a cronjob.

kebechet/managers/manager.py
```diff
@@ -63,7 +63,7 @@ def __init__(
 
         self.service = service
         self.project = service.get_project(namespace=self.owner, repo=self.repo_name)
-        self._repo = None
+        self._repo: git.Repo
```
harshad16, line 66: Shouldn't the change include `=` instead of `:` ?

harshad16 (changes requested):
Any repo is included to kebechet, by installing kebhut app and including `.thoth.yaml` in the repo.
There are no changes required in the kebechet repo.

config/thoth.yaml
Gregory-Pereira: > Any repo is included to kebechet, by installing kebhut app and including `.thoth.yaml` in the repo. There are no changes required in the kebechet repo.

So if we don't need this to install with kebechet I should just close the whole PR correct?

kebechet/managers/manager.py
Gregory-Pereira: Rather than assignment this was supposed to be type defintion, which is why a colon is being used. Otherwise the pylance tests threw errors.

goern:
what is the status of this? can we release it with the next milestone?

config/thoth.yaml
harshad16: yes this change is not needed, there seems to be another change in this PR.
if that is also not needed, this can be closed.

kebechet/managers/manager.py
harshad16: I see, then can we assign the default value as well.
`self._repo: git.Repo = None`

harshad16:
> what is the status of this? can we release it with the next milestone?

It is under Review.
The changes are not for release, it is a simple type fix.
This a minor PR.

goern:
@Gregory-Pereira is this good to approve?

harshad16 (approved):
/lgtm
/approve

Issue #1059, opened 2022-06-08
[1pt] Delete the branch if the pr is merged and tag is created.


harshad16:
As a kebechet operator, I would like to have kebechet clean the branch once it has fulfilled its purpose so that it would make the repository maintainers' life easier.

**Acceptance criteria**
- [ ] Delete the branch if the pr is merged and the tag is created.
- [ ] Don't delete the branch if pr is pending or closed.

harshad16:
@KPostOffice @Gkrumbach07 can please give your ack on this issue, if you feel this is an apt request.

KPostOffice:
ack. This is doable and a usefule feature I think

codificat:
#1065 addressed this, so 
/close

PR #1065, opened 2022-06-13
Cleanup branch after  PR is merged and tag is created

shreekarSS:
Fixes cleaning up remote branch after pr is merged and tag is created

https://github.com/thoth-station/kebechet/issues/1059
…

As part of the cleanup, it checks if pr has been merged, creates the tag, and deletes the branch on the remote.

### DIFF
--- a/kebechet/managers/version/version.py
+++ b/kebechet/managers/version/version.py
@@ -226,8 +226,27 @@ class VersionManager(ManagerBase):
                         ["git", "push", "origin", f"refs/tags/{tag_version}"]
                     )
                 )
-            return
+                try:
+                    # Branch name is same as tag version
+                    _LOGGER.info(f"Deleting branch {tag_version}")
+                    repo.git.execute(
+                        [
+                            "git",
+                            "push",
+                            "origin",
+                            "-d",
+                            f"refs/heads/{tag_version}",
+                        ]
+                    )
 
+                except Exception:
+                    _LOGGER.exception(f"Failed to delete branch {tag_version}")
+                    self.pr_comment(
+                        utils._pr_id_from_webhook(self.parsed_payload),
+                        body=f"Failed to delete branch {tag_version} , due to permissions issue",
+                    )
+
+            return
         if (
             pr_releases
             and self.parsed_payload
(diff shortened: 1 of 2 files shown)


kebechet/managers/version/version.py
```diff
@@ -226,8 +226,29 @@ def run(  # type: ignore
                         ["git", "push", "origin", f"refs/tags/{tag_version}"]
                     )
                 )
-            return
+                try:
+                    # Branch name is same as tag version
+                    branch_to_delete = tag_version
```
harshad16, line 231: can directly use the tag_version ? instead of creating another variable?

kebechet/managers/version/version.py
```diff
@@ -226,8 +226,29 @@ def run(  # type: ignore
...
+                    )
+
+                except Exception:
+                    _LOGGER.exception(f"Failed to delete branch {branch_to_delete}")
 
+                    self.project.create_issue(
+                        title="Failed to cleanup the release version branch",
+                        body="Failed to delete branch by kebechet bot after tag creation",
+                        labels=self.labels,
+                    )
```
harshad16, line 250: As we are in the PR, let us just comment on the PR, that the bot cant delete the branch due to lack of permissions.

harshad16 (changes requested):
lgtm
great work
just few comments

kebechet/managers/version/version.py
harshad16: yes the user would be able to see this info.
comments are allowed on closed PRs.

kebechet/managers/version/version.py
shreekarSS: ACK!

kebechet/managers/version/version.py
harshad16: It sends a notification to all the participants of the PR.(reviewers, approvers, repo maintainers, subscribers).
The issue in my opinion would need more handle like, if we open an issue in one repo, then we check if there is an existing one and the same kinda scenario.
also if the user doesn't want to give permission then this way there would only be comments. In case of an issue, we don't have a feedback loop.

harshad16 (commented):
/lgtm
thanks :100: 
PTAL @KPostOffice 

@shreekarSS  can we please rebase the commit to 1.

shreekarSS:
> /lgtm thanks 100 PTAL @KPostOffice
> 
> @shreekarSS can we please rebase the commit to 1.

ack

harshad16 (approved):

PR #1081, opened 2022-07-08
Issue: Kebechet was picking approvers from it's own OWNERS file

shreekarSS:
Fix: To pick approvers from CLONED repo

https://github.com/thoth-station/ps-nlp/issues/151

https://github.com/shreekarSS/ps-nlp/issues/5

## Description

Kebechet will pull the approvers list from the cloned repository rather than its own OWNERS file.

### DIFF
--- a/kebechet/managers/thoth_advise/thoth_advise.py
+++ b/kebechet/managers/thoth_advise/thoth_advise.py
@@ -230,8 +230,7 @@ class ThothAdviseManager(ManagerBase):
 
     def _get_users_with_permission(self) -> typing.List[str]:
         try:
-            with open("OWNERS", "r") as owners_file:
-                owners = yaml.safe_load(owners_file)
+            owners = yaml.safe_load(self.project.get_file_content("OWNERS"))
             permitted_users = list(map(str, owners.get("approvers") or [])) + [APP_NAME]
         except FileNotFoundError:
             permitted_users = list(self.project.who_can_merge_pr()) + [APP_NAME]


kebechet/managers/thoth_advise/thoth_advise.py
```diff
@@ -229,12 +229,15 @@ def _act_on_advise_error(self, adv_results: dict):
             self._tracking_issue.comment(comment)
 
     def _get_users_with_permission(self) -> typing.List[str]:
-        try:
```
harshad16, line 232: seems like we are missing this bit:
```
with cloned_repo(self, self.project.default_branch, depth=1) as repo:
                self.repo = repo
```
please include this bit before try catch block and test it

harshad16 (changes requested):
checking ownership is a good thing, so lets keep it and fix issue regarding it.

kebechet/managers/thoth_advise/thoth_advise.py
KPostOffice: Yeah, that looks to be the issue. Even better, you can use `self.project.get_file_content("OWNERS")` so that only one file gets downloaded.

kebechet/managers/thoth_advise/thoth_advise.py
```diff
@@ -229,9 +229,18 @@ def _act_on_advise_error(self, adv_results: dict):
...
+        # try:
+        #     with open("OWNERS", "r") as owners_file:
+        #
+        #         owners = yaml.safe_load(owners_file)
+        #         _LOGGER.info(owners)
+        #     permitted_users = list(map(str, owners.get("approvers") or [])) + [APP_NAME]
+        # except FileNotFoundError:
+        #     permitted_users = list(self.project.who_can_merge_pr()) + [APP_NAME]
+        # permitted_users = list(self.project.who_can_merge_pr()) + [APP_NAME]
+
```
harshad16, line 241: please comment properly or remove it or adjust on this itself.

harshad16 (changes requested):
just one question

harshad16 (approved):
/lgtm
/approve
thanks :100:

Issue #1072, opened 2022-06-21
Info manager fails to provide info


codificat:
## Bug description

I requested Kebechet to provide information on a stack, and it provided an error instead.

### Steps to Reproduce
1. Have Kebechet configured on your repo
2. Open an issue in the repo, with title: *Kebechet info*
3. Wait for a comment from the bot in the issue

Example: https://github.com/the-shadowmen/nlp-first-steps/issues/3#issuecomment-1152661245

### Actual behavior

The example linked above includes this in the *Dependency graph* section:

``` text
virtualenv: error: argument dest: the destination . is not write-able at /home/user/.local/share/virtualenvs

Failed to create virtual environment.
```

After the comment, the issue gets closed by the bot.

### Expected behavior

- There should be no error
- If there is an error, provide next steps (even if only pointing the user to the support repo)

### Environment information

From the example linked above:

``` text
Kebechet version: 1.9.0
Python version: 3.8.12
Platform: Linux-4.18.0-305.19.1.el8_4.x86_64-x86_64-with-glibc2.2.5
pipenv version: pipenv, version 2020.11.15
```

Gkrumbach07:
**suggested actions**
- [ ] verify that failure is from dependency issues
- [ ] install cmake on kebechet image and pipfile.lock that has error

**story points** 5

/triage accepted

shreekarSS:
I tried the following steps to fix:
1) Update piplock with cmake dependency
2) Update Dockerfile with cmake dependency

Tested locally, got expected info https://github.com/shreekarSS/nlp-first-steps/issues/2.
Yet to test docker image with  the updates.

PR #1100, opened 2022-07-29
Update pipfile,Dockerfile and pipfile.lock files with cmake dependency

shreekarSS:

Fixes: https://github.com/thoth-station/kebechet/issues/1072


KPostOffice:
You shouldn't need to add cmake to the Pipfile. I think just having the cmake library installed in the container should be enough

PR #1109, opened 2022-08-04
call sync when generating dependency graph

KPostOffice:
otherwise arbitrary sub-dependencies will be found to satisfy the requirements

I noticed install was being called here which takes longer but also gets an arbitrary dependency graph. Sync will install deps directly from the lockfile


codificat:
good find!
/approve

Issue #1107, opened 2022-08-02
kebechet version release manager is failing


harshad16:
## Bug description

Kebechet is unable to resolve the version update. The update manager goes over a list of files to find the version and update it.
The version manager should find these files in the repo it working on. however, it feels like the version manager checks on the entire container image. 

https://github.com/thoth-station/adviser/issues/2370#issuecomment-1201852695
```
Details:

File path: ./.venv/lib/python3.8/site-packages/pkg_resources/_vendor/packaging/__about__.py
Line number: 20
Old version: 21.3
Issue: Wrong version specifier found in sources: 21.3``
```

### Steps to Reproduce
1. Run kebechet version update on adviser repo.
2. See error

### Actual behavior
the version manager checks for version file in entire container image.

### Expected behavior
version manager should only check repo directory for version file.

### Environment information

Kebechet version: 1.10.2
Python version: 3.8.12
Platform: Linux-4.18.0-305.19.1.el8_4.x86_64-x86_64-with-glibc2.2.5
pipenv version: pipenv, version 2020.11.15

### Additional context
My suggestion to solve this issue would be to check if this [function](https://github.com/thoth-station/kebechet/blob/0f907e23ae1dfa4d828c519a43d80871188bfacd/kebechet/managers/version/release_triggers.py#L92) is being executed on the cloned repo.

harshad16:
Related-to: https://github.com/thoth-station/workflow-helpers/issues/408#issuecomment-1203151660

KPostOffice:
It is being called in the cloned repo. It might be best to just ignore files and directories that start with a "."

KPostOffice:
Though the venv shouldn't be in the repository

KPostOffice:
I think the issue lies in `cwd` it is failing silently and cloning directly into the root of the repo

KPostOffice:
Issue ended up being the venv was being created inside the repo directory from https://github.com/thoth-station/kebechet/blob/86e803e3816213b31d454230bd64ebad466c931f/kebechet/managers/update/update.py#L802

To fix I have decided to ignore hidden files and directories during the walk

PR #1110, opened 2022-08-04
remove hidden files during os walk

KPostOffice:

fixes: https://github.com/thoth-station/kebechet/issues/1107

- [ ] Yes
- [x] No

ignore hidden files and directories during walk

`.venv` dir can be created in cloned directory by update manager which messes with version discovery and writing

### DIFF
--- a/kebechet/managers/version/release_triggers.py
+++ b/kebechet/managers/version/release_triggers.py
@@ -94,7 +94,12 @@ class BaseTrigger:
     ) -> List[Tuple[str, str, str]]:
         """Walk through the directory structure and try to adjust version identifier in sources."""
         adjusted = []
-        for root, _, files in os.walk("./"):
+        for root, dirs, files in os.walk("./"):
+            dirs[:] = [
+                d for d in dirs if not d[0] == "."
+            ]  # remove hidden dirs (done in place to remove from walk)
+            files = [f for f in files if not f[0] == "."]  # remove hidden files
+
             for file_name in files:
                 if file_name in (
                     "setup.py",


kebechet/managers/version/release_triggers.py
```diff
@@ -94,7 +94,12 @@ def adjust_version_in_sources(
     ) -> List[Tuple[str, str, str]]:
         """Walk through the directory structure and try to adjust version identifier in sources."""
         adjusted = []
-        for root, _, files in os.walk("./"):
+        for root, dirs, files in os.walk("./"):
+            dirs[:] = [
+                d for d in dirs if not d[0] == "."
+            ]  # remove hidden dirs (done in place to remove from walk)
+            files = [f for f in files if not f[0] == "."]  # remove hidden files
```
harshad16, line 101: As os.walk would explicity walk over each file sub-directory after sub-directory
maybe we could just check the root ?
```suggestion
     if not root.startswith("./.")             
```
wdyt about this ?

harshad16 (changes requested):
great find :tada: :medal_military: 
just one comment:
we are any not using the dirs, only the files search, so this would files list might still have the hidden dirs, it seems

kebechet/managers/version/release_triggers.py
KPostOffice: The in place redefinition of dirs keeps walk from iterating over hidden directories and finding those files

harshad16: thanks.

harshad16 (approved):
/lgtm
/approve
thanks

Issue #1125, opened 2022-08-31
[3pt] Allow local and git package refereces in update manager


codificat:
## Problem statement

When I use Kebechet's update manager, which uses `pipenv` to resolve dependencies, it refuses to invoke `pipenv lock` if my `Pipfile` contains references to local paths or pointers to git.

For example, see https://github.com/AICoE/sefkhet-abwy/issues/192 where Kebechet says:

> Kebechet cannot support maintaining this application as it contain's git version of packages.

This is a bit frustrating, because `pipenv` would actually handle these. 

Continuing the example above, the *sefkhet-abwy* repository's `Pipfile` contains [a git referenced package](https://github.com/AICoE/sefkhet-abwy/blob/433951fa980373ba7ad8df8c340a658ae90296dd/Pipfile#L18). But `pipenv lock` works:

```
$ pipenv lock
Locking [dev-packages] dependencies...
Building requirements...
Resolving dependencies...
✔ Success! 
Locking [packages] dependencies...
Building requirements...
Resolving dependencies...
✔ Success! 
Updated Pipfile.lock (e1d779)!

$ git status
On branch master
Your branch is up to date with 'upstream/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Pipfile.lock

no changes added to commit (use "git add" and/or "git commit -a")

$ pipenv graph | grep -A3 ChatterBot
ChatterBot==1.1.0a7
  - mathparse [required: >=0.1,<0.2, installed: 0.1.2]
  - python-dateutil [required: >=2.8,<2.9, installed: 2.8.2]
    - six [required: >=1.5, installed: 1.16.0]
```

## High-level Goals

As the update manager effectively uses `pipenv lock` to update dependencies, the goal of this request is to let the manager run in scenarios where `pipenv lock` runs.

## Proposal description

Do not block execution of `pipenv` resolution.

### Alternatives

The alternative of leaving the restriction (no git, no local) in place is consistent with the requirements of the thoth-advise manager.

However, as `pipenv` itself does not impose the restrictions that thoth advise does, the current behaviour makes it inconsistent with `pipenv`.

### Additional context

I did a quick search on when the check for git references was introduced, and AFAICT this was added in #820 to address https://github.com/thoth-station/support/issues/20

## Acceptance Criteria

[...]

PR #1140, opened 2022-09-14
Remove checks for git and local path from pipfile

shreekarSS:

Fixes: https://github.com/thoth-station/kebechet/issues/1125
…

### DIFF
--- a/kebechet/managers/update/update.py
+++ b/kebechet/managers/update/update.py
@@ -45,7 +45,6 @@ from .messages import (
     ISSUE_INITIAL_LOCK,
     ISSUE_NO_DEPENDENCY_MANAGEMENT,
     ISSUE_PIPENV_UPDATE_ALL,
-    ISSUE_UNSUPPORTED_PACKAGE,
     UNINIT_OVERLAY_DIR_BODY,
     UPDATE_MESSAGE_BODY,
     CLOSE_MANUAL_ISSUE_COMMENT_PR,
@@ -73,10 +72,6 @@ _ISSUE_REPLICATE_ENV_NAME = (
 )
 _ISSUE_NO_DEPENDENCY_NAME = (
     "No dependency management found for the {env_name} environment"
-)
-_ISSUE_UNSUPPORTED_PACKAGE = (
-    "Application cannot be managed by Kebechet due to it containing an unsupported package "
-    "location in {env_name} environment."
 )
 _ISSUE_MANUAL_UPDATE = "Kebechet update"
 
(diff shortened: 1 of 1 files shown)


KPostOffice:
To give some context. @shreekarSS and I talked a bit about this and realized that the result from the function raising the issues in git repositories wasn't being used. This was the only reference to the code in the repo, so we decided it could be removed.  The return statement to the get direct dependency version so that when old dependencies and new dependencies are calculated they don't throw error and instead report no differences in reported version.

kebechet/managers/update/update.py
```diff
@@ -235,62 +235,6 @@ def _get_direct_dependencies_requirements(req_dev: bool) -> set:
...
-            # TODO: open a PR to fix this
-            raise DependencyManagementError(
-                f"Failed to load Pipfile.lock file: {str(exc)}"
-            ) from exc
-
-        result = {}
-
-        for package_name, package_info in pipfile_lock_content["default"].items():
-            if "git" in package_info:
-                self._create_unsupported_package_issue(package_name, "git")
```
codificat, line 253: I think the definition of this method should also be removed with the proposed change, as it is only used within the removed code

https://github.com/thoth-station/kebechet/blob/a7c2f9d6c207cb3740b268fd2502c0bdb6c15bdf/kebechet/managers/update/update.py#L294-L327)

kebechet/managers/update/update.py
```diff
@@ -235,62 +235,6 @@ def _get_direct_dependencies_requirements(req_dev: bool) -> set:
...
-                    "Failed to find version for package that uses local installation."
-                )
-            result[package_name.lower()] = {
-                "dev": False,
-                "version": package_info["version"][len("==") :],
-            }
-        # Close git as a source issues.
-
-        issue = self.get_issue_by_title(
-            _ISSUE_UNSUPPORTED_PACKAGE.format(env_name=self.runtime_environment)
```
codificat, line 285: The definition of that message should also be removed if it is not used anymore

https://github.com/thoth-station/kebechet/blob/a7c2f9d6c207cb3740b268fd2502c0bdb6c15bdf/kebechet/managers/update/update.py#L77-L80

codificat (changes requested):
Thanks for this!

I believe there are a couple of additional code snippets that should be cleaned up as part of this change

kebechet/managers/update/update.py
shreekarSS: ack, thanks

kebechet/managers/update/update.py
shreekarSS: ack, thanks

codificat:
/lgtm
/approve

Thanks!

PR #1146, opened 2022-09-21
Create function which respects gh body limits

KPostOffice:
add function which allows for specifying required and optional info. Optional info will be truncated if body is > gh body limit

fixes: https://github.com/thoth-station/kebechet/issues/1122

- [ ] Yes
- [x] No


KPostOffice:
This should be good 2 go, but I would appreciate a review.

harshad16 (commented):
Seems like a great solution for the problem
/lgtm

concerns for long term: 
- if the req_info can be in a greater len than limit, we arent checking that.
- Even though we are trying to keep it under the limit, the limit seems to be still large comment
maybe we should cut it into half, for readability purposes.

let's get this in and see if we have a readability issue.

KPostOffice:
> Seems like a great solution for the problem /lgtm
> 
> concerns for long term:
> 
> * if the req_info can be in a greater len than limit, we arent checking that.
> * Even though we are trying to keep it under the limit, the limit seems to be still large comment
>   maybe we should cut it into half, for readability purposes.
> 
> let's get this in and see if we have a readability issue.

1. hopefully no one sees an entire back trace as "required" for a comment.
2. often the info that is longest is hidden by a "read more"

Issue #1158, opened 2022-10-18
[3pt] Support pyproject.toml and PEP 566 standards in Kebechet version manager


Gkrumbach07:
Support pyproject.toml and PEP 566 standards in Kebechet version manager

**Acceptance Criteria**
- [ ] Allow the version manager to manage the version in pyproject.toml based on PEP standard.
https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata
https://github.com/thoth-station/kebechet/blob/285c57cc12dc011ecb6374b4cd4f76abf6fd31ac/kebechet/managers/version/release_triggers.py#L104

PR #1163, opened 2022-10-26
Add pyproject.toml to maintain version from Kebechet's version manager

shreekarSS:

Fixes: https://github.com/thoth-station/kebechet/issues/1158

…

### DIFF
--- a/kebechet/managers/version/messages.py
+++ b/kebechet/managers/version/messages.py
@@ -43,11 +43,13 @@ ISSUE_BODY_NO_VERSION_IDENTIFIER_FOUND = """
 Automated version release could not be completed.
 
 Kebechet version manager expects a file with one of the following names: `["setup.py", "__init__.py", "__about__.py",
-"version.py", "app.py", "wsgi.py"]` to contain the line:
+"version.py", "app.py", "wsgi.py", "pyproject.toml"]` to contain the line:
 
 ```
 ...
-__version__ = X.Y.Z
+version = X.Y.Z  (in pyproject.toml project section)
+or
+__version__ = X.Y.Z (in the rest of the options)
 ...
 ```
 
--- a/kebechet/managers/version/release_triggers.py
+++ b/kebechet/managers/version/release_triggers.py
@@ -16,12 +16,12 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 """Release triggers which result in new version released on GitHub."""
-
 from typing import List, Optional, Any, Dict, Tuple
 import logging
 import os
 import json
 
+import toml
 from ogr.abstract import Issue, PullRequest
 from kebechet.utils import get_issue_by_title
 from kebechet.managers.manager import ManagerBase
@@ -40,13 +40,38 @@ _LOGGER = logging.getLogger(__name__)
 class BaseTrigger:
     """Base class for triggers of new version releases."""
 
+    def adjust_version_toml_file(self, file_path: str) -> Optional[tuple]:
+        """Adjust version in the toml file, return signalizes whether the return value indicates change in file."""
+        with open(file_path, "r") as f:
+            toml_content = toml.loads(f.read())
+        project = toml_content.get("project", None)
+        old_version = project.get("version")
+        if old_version:
+            try:
+                new_version = self.get_new_version(old_version)
+            except VersionError as e:
+                raise VersionError(
+                    UNABLE_TO_UPDATE_VERSION_ERROR.format(
+                        file_path=file_path,
+                        line_num=1,
+                        old_version=old_version,
+                        reason=str(e),
+                        environment_details=ManagerBase.get_environment_details(),
+                    )
+                ) from e
+            _LOGGER.info("Computed new version: %s", new_version)
+            toml_content["project"]["version"] = new_version
+            with open(file_path, "w") as output_file:
+                output_file.write(toml.dumps(toml_content))
+            return new_version, old_version
+        return None
+
     def adjust_version_file(self, file_path: str) -> Optional[tuple]:
         """Adjust version in the given file, return signalizes whether the return value indicates change in file."""
-        with open(file_path, "r") as input_file:
-            content = input_file.read().splitlines()
-
         changed = False
         old_version = None
+        with open(file_path, "r") as input_file:
+            content = input_file.read().splitlines()
         for idx, line in enumerate(content):
             if line.startswith("__version__ = "):
                 parts = line.split(" = ", maxsplit=1)
@@ -72,12 +97,9 @@ class BaseTrigger:
                             environment_details=ManagerBase.get_environment_details(),
                         )
                     ) from e
-
                 _LOGGER.info("Computed new version: %s", new_version)
-
                 content[idx] = f'__version__ = "{new_version}"'
                 changed = True
-
         if not changed:
             return None
 
@@ -98,22 +120,29 @@ class BaseTrigger:
             dirs[:] = [
                 d for d in dirs if not d[0] == "."
             ]  # remove hidden dirs (done in place to remove from walk)
+            if "tests" in dirs:
+                dirs.remove("tests")
             files = [f for f in files if not f[0] == "."]  # remove hidden files
-
+            files_dict = {
+                "setup.py": self.adjust_version_file,
+                "__init__.py": self.adjust_version_file,
+                "__about__.py": self.adjust_version_file,
+                "version.py": self.adjust_version_file,
+                "app.py": self.adjust_version_file,
+                "wsgi.py": self.adjust_version_file,
+                "pyproject.toml": self.adjust_version_toml_file,
+            }
             for file_name in files:
-                if file_name in (
-                    "setup.py",
-                    "__init__.py",
-                    "__about__.py",
-                    "version.py",
-                    "app.py",
-                    "wsgi.py",
-                ):
-                    file_path = os.path.join(root, file_name)
-                    adjusted_version = self.adjust_version_file(file_path)
+                if file_name in files_dict:
+                    func_to_call = files_dict[file_name]
+                    adjusted_version = func_to_call(os.path.join(root, file_name))
                     if adjusted_version:
                         adjusted.append(
-                            (file_path, adjusted_version[0], adjusted_version[1])
+                            (
+                                os.path.join(root, file_name),
+                                adjusted_version[0],
+                                adjusted_version[1],
+                            )
                         )
         return adjusted
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,7 @@ repos:
     hooks:
       - id: black
 
-  - repo: https://gitlab.com/PyCQA/flake8
+  - repo: https://github.com/PyCQA/flake8
     rev: '3.9.2'
     hooks:
       - id: flake8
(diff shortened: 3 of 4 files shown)


kebechet/managers/version/release_triggers.py
```diff
@@ -108,6 +126,7 @@ def adjust_version_in_sources(
                     "version.py",
                     "app.py",
                     "wsgi.py",
```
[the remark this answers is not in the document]
shreekarSS: @KPostOffice I'm thinking about using dict of {file_name : function} , fetch the function based on file_name and call

kebechet/managers/version/release_triggers.py
```diff
@@ -72,20 +102,16 @@ def adjust_version_file(self, file_path: str) -> Optional[tuple]:
...
-            output_file.write("\n".join(content))
-            # Add new line at the of file explicitly.
-            output_file.write("\n")
+            if not changed:
+                return None
+            # Apply changes.
+            with open(file_path, "w") as output_file:
+                output_file.write("\n".join(content))
+                # Add new line at the of file explicitly.
+                output_file.write("\n")
```
harshad16, line 114: This part should be outside the for loop, right?
or else as soon as first line isn't __version__ it would return None.
we should trace to the whole loop before we check for changed.

harshad16 (changes requested):
Good approach.

kebechet/managers/version/messages.py
```diff
@@ -43,11 +43,13 @@
...
 
 Kebechet version manager expects a file with one of the following names: `["setup.py", "__init__.py", "__about__.py",
-"version.py", "app.py", "wsgi.py"]` to contain the line:
+"version.py", "app.py", "wsgi.py", "pyproject.toml"]` to contain the line:
 
 ```
 ...
 __version__ = X.Y.Z
+or
+version = X.Y.Z
```
harshad16, line 52: maybe it would be good to specific file like

```suggestion
version = X.Y.Z  (in pyproject.toml)
or
__version__ = X.Y.Z (in the rest of the options)
```

kebechet/managers/version/release_triggers.py
```diff
@@ -98,19 +123,24 @@ def adjust_version_in_sources(
...
-                    "version.py",
-                    "app.py",
-                    "wsgi.py",
-                ):
-                    file_path = os.path.join(root, file_name)
-                    adjusted_version = self.adjust_version_file(file_path)
+                if file_name in files_dict:
+                    file_path: str = os.path.join(root, file_name)
+                    func_to_call = files_dict.get(file_name)
+                    adjusted_version = func_to_call(file_path)
```
harshad16, line 143:
```suggestion
                    func_to_call: Any = files_dict[file_name]
                    adjusted_version = func_to_call(os.path.join(root, file_name))
```
will this work ?

harshad16 (changes requested):
some more suggestions

kebechet/managers/version/release_triggers.py
shreekarSS: My bad, while refactoring, I must have tab- pushed it . Thanks for the review.

kebechet/managers/version/release_triggers.py
shreekarSS: ack, worked

harshad16 (commented):
/lgtm
PTAL @KPostOffice

harshad16 (commented):
One more note
update the docs as well
https://github.com/thoth-station/kebechet/blob/master/docs/managers/version.rst

VannTen:
I have a question:
How do we handle version specified not in by the [project] section, but provided by the build backend ?
I'm thinking specifically of setup like:
```toml
[project]
dynamic = ["version"]

[tool.setuptools-scm]
enabled = true
```

or
```toml
[project]
dynamic = ["version"]

[tool.setuptools.dynamic]
version = {attr = "thoth.document_sync.__version__"}
```
(I'm not sure it's possible in a general way, but in that case we might want to state that in the docs)
```

**Comment by harshad16** [comment]
That is a valid point, we should check/address if the version is not in the project section.

File: kebechet/managers/version/messages.py [hunk 5]
```diff
@@ -43,11 +43,13 @@
 Automated version release could not be completed.
 
 Kebechet version manager expects a file with one of the following names: `["setup.py", "__init__.py", "__about__.py",
-"version.py", "app.py", "wsgi.py"]` to contain the line:
+"version.py", "app.py", "wsgi.py", "pyproject.toml"]` to contain the line:
 
 ```
 ...
-__version__ = X.Y.Z
+version = X.Y.Z  (in pyproject.toml)
```
harshad16, line 50:
```suggestion
version = X.Y.Z  (in pyproject.toml project section)
```

shreekarSS: Ack, done

harshad16 (commented):
/lgtm

PTAL @KPostOffice

KPostOffice:
https://github.com/python-poetry/poetry/issues/144

An interesting thread about why "version" should never be put in source code and should instead be derived from package metadata when needed.

The basic argument here is that __version__ should be set from package metadata, not the other way around.

KPostOffice:
I'm not sure what this should mean for this PR, just a rabbit hole that I fell down when looking into Max's concern.

One way to just fix the issue raised is to recognize that json if `version` is a dictionary then it won't hold a proper string and to return early. If it is specified in a different file then the version manager should update it automatically already so throwing an exception would be wrong in this case.

harshad16:
I believe we are addressing the required pattern with PR.
we are only updating the pyproject.toml version if it is the project section, or else it wouldn't be picked for updated.
and we are describing this in our docs as well.

KPostOffice:
> I believe we are addressing the required pattern with PR. we are only updating the pyproject.toml version if it is the project section, or else it wouldn't be picked for updated. and we are describing this in our docs as well.

Oh yeah, I see what you mean

kebechet/managers/version/release_triggers.py
```diff
@@ -40,13 +40,42 @@
...
 
+    def adjust_version_toml_file(self, file_path: str) -> Optional[tuple]:
+        """Adjust version in the toml file, return signalizes whether the return value indicates change in file."""
+        changed = False
+        with open(file_path, "r") as f:
+            toml_content = toml.loads(f.read())
+        project = toml_content.get("project", None)
+        old_version = project.get("version", None)
+        try:
+            new_version = self.get_new_version(old_version)
```
KPostOffice, line 51: Old version must be a string. Passing `None` here will throw an error.

kebechet/managers/version/release_triggers.py
```diff
@@ -40,13 +40,42 @@
...
+                UNABLE_TO_UPDATE_VERSION_ERROR.format(
+                    file_path=file_path,
+                    line_num=1,
+                    old_version=old_version,
+                    reason=str(e),
+                    environment_details=ManagerBase.get_environment_details(),
+                )
+            ) from e
+        _LOGGER.info("Computed new version: %s", new_version)
+        if old_version:
```
KPostOffice, line 63: This check should occur above and just return if `version` exists in `pyproject.toml`

kebechet/managers/version/release_triggers.py
```diff
@@ -40,13 +40,42 @@
...
+            toml_content["project"]["version"] = new_version
+            changed = True
+        if not changed:
```
KPostOffice, line 66: Having `changed` and this conditional will be unnecessary if old_version is checked to not be `None` above

KPostOffice (commented):
Looks good. Just a few small changes to the new function

kebechet/managers/version/release_triggers.py
shreekarSS: Ack

kebechet/managers/version/release_triggers.py
shreekarSS: moved it above, thanks

kebechet/managers/version/release_triggers.py
```diff
@@ -40,13 +40,38 @@
...
+                        old_version=old_version,
+                        reason=str(e),
+                        environment_details=ManagerBase.get_environment_details(),
+                    )
+                ) from e
+        _LOGGER.info("Computed new version: %s", new_version)
+        toml_content["project"]["version"] = new_version
+        new_content = toml.dumps(toml_content)
+        with open(file_path, "w") as output_file:
+            output_file.write(new_content)
```
KPostOffice, line 66: Here you can directly do `toml.dump(toml_content, output_file)`, no need for an intermediate string representation.

kebechet/managers/version/release_triggers.py
```diff
@@ -98,22 +120,29 @@ def adjust_version_in_sources(
...
-                    "__init__.py",
-                    "__about__.py",
-                    "version.py",
-                    "app.py",
-                    "wsgi.py",
-                ):
-                    file_path = os.path.join(root, file_name)
-                    adjusted_version = self.adjust_version_file(file_path)
+                if file_name in files_dict:
+                    func_to_call: Any = files_dict[file_name]
```
KPostOffice, line 137: could use `Callable[[str], Optional[Tuple[str, str]]]` as a type hint here. If it gives any errors, we can just leave it as callable.

KPostOffice (commented):
Just a couple of small comments and then it should be good I think

kebechet/managers/version/release_triggers.py
shreekarSS: Pre commit failed with : error: Incompatible types in assignment (expression has type "Callable[[str], Optional[Tuple[Any, ...]]]", variable has type "Callable[[str], Optional[Tuple[str, str]]]")

kebechet/managers/version/release_triggers.py
shreekarSS: done

kebechet/managers/version/release_triggers.py
KPostOffice: That's because of the function signatures of the called functions could be more descriptive too, so you would need to change this line as well as the the function signatures.

kebechet/managers/version/release_triggers.py
```diff
@@ -40,13 +40,37 @@
...
+                raise VersionError(
+                    UNABLE_TO_UPDATE_VERSION_ERROR.format(
+                        file_path=file_path,
+                        line_num=1,
+                        old_version=old_version,
+                        reason=str(e),
+                        environment_details=ManagerBase.get_environment_details(),
+                    )
+                ) from e
+        _LOGGER.info("Computed new version: %s", new_version)
```
KPostOffice, line 62: These need to be shifted over so that they are inside the conditional I'm pretty sure.

shreekarSS: ack, done

kebechet/managers/version/release_triggers.py
```diff
@@ -40,13 +40,37 @@
...
+                        old_version=old_version,
+                        reason=str(e),
+                        environment_details=ManagerBase.get_environment_details(),
+                    )
+                ) from e
+            _LOGGER.info("Computed new version: %s", new_version)
+            toml_content["project"]["version"] = new_version
+            with open(file_path, "w") as output_file:
+                output_file.write(toml.dumps(toml_content))
+        return new_version, old_version
```
KPostOffice, line 66: This also needs to be tabbed over. If there is no version, it should just return None

shreekarSS: ack

harshad16 (commented):
/unhold
/lgtm

thanks :+1:

Issue #1067, opened 2022-06-21
[2pt] Add signed-off-by in commit messages


codificat:
## Problem statement

When Kebechet sends e.g. a PR to update dependencies, the commit message(s) do not contain a `Signed-off-by:` line.

This can prevent [DCO](https://developercertificate.org/) checks from working.

Example: https://github.com/os-climate/data-platform-demo/pull/41/checks?check_run_id=6642425305

## High-level Goals

To have all of Kebechet's commits `Signed-off-by` the bot.

## Acceptance Criteria

- [ ] for every commit that Kebechet generates, the commit message always includes a line `Signed-off-by: Kebechet <kebechet-s@email.example.com>`

Gkrumbach07:
**story points** 2

PR #1165, opened 2022-11-02
add signing to kebechet commits

KPostOffice:
Signed-off-by: Kevin Postlethwait <kpostlet@redhat.com>

fixes: #1067

- [ ] Yes
- [x] No

adding the signoff flags to commits

see:
https://gitpython.readthedocs.io/en/0.2/tutorial.html#using-git-directly
https://docs.pi-hole.net/guides/github/how-to-signoff/

### DIFF
--- a/kebechet/managers/config_initializer/config_initializer.py
+++ b/kebechet/managers/config_initializer/config_initializer.py
@@ -65,6 +65,7 @@ class ConfigInitializer(ManagerBase):
         thoth_config = pkg_resources.read_text(resources, "simple.thoth.yaml")
 
         with cloned_repo(self, depth=1, branch=self.project.default_branch) as repo:
+            self.repo = repo
             prs = self.get_prs_by_branch(_BRANCH_NAME, status=PRStatus.all)
             if len(prs) > 0:
                 _LOGGER.debug("PR initializing .thoth.yaml already exists skipping...")
@@ -72,9 +73,11 @@ class ConfigInitializer(ManagerBase):
             repo.git.checkout("HEAD", b=_BRANCH_NAME)
             with open(".thoth.yaml", "w+") as f:
                 f.write(thoth_config)
-            repo.index.add([".thoth.yaml"])
-            repo.index.commit("Initialize .thoth.yaml with basic configuration")
-            repo.remote().push(_BRANCH_NAME)
+            self._git_commit_push(
+                "Initialize .thoth.yaml with basic configuration",
+                _BRANCH_NAME,
+                [".thoth.yaml"],
+            )
             self.create_pr(
                 title="Thoth Configuration Initialization",
                 body=_PR_BODY,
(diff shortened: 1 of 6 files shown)


kebechet/managers/config_initializer/config_initializer.py
```diff
@@ -73,7 +73,9 @@ def run(self) -> typing.Optional[dict]:  # type: ignore
             with open(".thoth.yaml", "w+") as f:
                 f.write(thoth_config)
             repo.index.add([".thoth.yaml"])
-            repo.index.commit("Initialize .thoth.yaml with basic configuration")
+            repo.index.commit(
+                "Initialize .thoth.yaml with basic configuration", signoff=True
```
codificat, line 77: The [commit method](https://gitpython.readthedocs.io/en/3.1.29/reference.html#git.index.base.IndexFile.commit) doesn't seem to understand a `signoff` argument:

```python
>>> idx.add('deleteme')
[(100644, e69de29bb2d1d6434b8b29ae775ad8c2e48c5391, 0, deleteme)]
>>> idx.commit("testing gitpython", signoff=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: commit() got an unexpected keyword argument 'signoff'
```

kebechet/managers/manager.py
```diff
@@ -170,6 +170,21 @@ def create_pr(self, title: str, body: str, source_branch: str, target_branch: st
...
 
+    def _git_commit_push(
+        self, commit_msg: str, branch_name: str, files: list, force_push: bool = False
+    ) -> None:
+        """Perform git push after adding files and giving a commit message."""
+        cur_branch = self.repo.active_branch
+        self.repo.git.checkout("HEAD", b=branch_name)
+        try:
+            self.repo.index.add(files)
+            self.repo.git.commit(f"-m {commit_msg}", "-s")
```
codificat, line 181: Everythnig after `-m` becomes the commit message, so this should be at least:
```suggestion
            self.repo.git.commit(f"-m{commit_msg}", "-s")
```

However, one more thing about this: if the `commit_msg` is multi-line, it should check that there is a blank line after the first one... It might be that all commits right now only contain a title, but should we check to make this future-proof?

codificat (changes requested):
Just one nit

kebechet/managers/manager.py
harshad16: The space is needed after `-m` . right ?
https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--mltmsggt

codificat: I tested it without the space and it works fine. With the space, the commit title starts with a space.

I guess it can be more clear if the change is this instead:
```suggestion
            self.repo.git.commit(f"--message={commit_msg}", "--signoff")
```

KPostOffice: Can do. I'll make the flags more explicit

Issue #1169, opened 2022-11-14
Facing issue when using kebechet for version update only 


harshad16:
## Bug description

We use kebechet for only version updates on non-python type code-base repositories.
when tried to make a new release using the kebechet version manager. we are facing following issues:

```
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.8/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "kebechet-cli", line 167, in cli_run_webhook
    run_webhook(payload=payload)
  File "/home/user/kebechet/kebechet_runners.py", line 83, in run_webhook
    run_url(url, service_type, parsed_payload)
  File "/home/user/kebechet/kebechet_runners.py", line 104, in run_url
    run(
  File "/home/user/kebechet/kebechet_runners.py", line 183, in run
    runtime_environments = [config.config["runtime_environments"][0]["name"]]
KeyError: 'runtime_environment
```

current kebechet version: `v1.11.0`
.thoth.yaml file used: https://github.com/thoth-station/thoth-application/blob/master/.thoth.yaml

### Steps to Reproduce
1. Try making a release for repositories with attached .thoth.yaml config
2. See error

harshad16:
Related-to: https://github.com/thoth-station/thoth-application/issues/2698

PR #1172, opened 2022-11-15
add default in case there are no runtime environments

KPostOffice:
some managers do not require python runtime environments, like the version manager. I changed to use an empty list as a default in this case. Any loops over runtime environments just won't run

Signed-off-by: Kevin <kpostlet@redhat.com>

fixes: #1169


harshad16 (approved):
/lgtm
/approve

thanks :+1:

Issue #1174, opened 2022-11-18
New patch release


harshad16:
Hey, Kebechet!

Create a new patch release, please.

harshad16:
```
2022-11-18 20:27:19,947   1 INFO     kebechet:219: Running manager 'version' for 'thoth-station/kebechet'
2022-11-18 20:27:26,355   1 DEBUG    kebechet.managers.manager:156: Issue No release maintainers stated for this repository not found, not closing.
2022-11-18 20:27:26,355   1 INFO     kebechet.managers.version.version:354: Found an issue #1174 which is a candidate for request of new version release: New patch release
2022-11-18 20:27:26,756   1 INFO     kebechet.managers.version.release_triggers:61: Old version found in sources: 1.11.0
2022-11-18 20:27:26,757   1 INFO     kebechet.managers.version.release_triggers:76: Computed new version: 1.11.1
2022-11-18 20:27:26,799   1 INFO     kebechet.managers.version.utils:129: Computing changelog for new release from version '1.11.0' to version '1.11.1'
2022-11-18 20:27:26,800   1 INFO     kebechet.managers.version.utils:146: Smart Log : False
2022-11-18 20:27:26,834   1 INFO     kebechet.managers.version.utils:197: Computed changelog has 5 entries
2022-11-18 20:27:26,837   1 INFO     kebechet.managers.version.utils:84: Adding changelog to the CHANGELOG.md file
2022-11-18 20:27:27,287   1 WARNING  git.remote:900: Error lines received while fetching: error: remote unpack failed: index-pack abnormal exit
error: failed to push some refs to 'https://github.com/thoth-station/kebechet'
2022-11-18 20:27:27,391   1 DEBUG    kebechet.managers.manager:110: Running pipenv command 'pipenv --version'
2022-11-18 20:27:29,558   1 ERROR    kebechet:237: An error occurred during run of manager {'configuration': {'assignees': ['sesheta'], 'labels': ['bot'], 'changelog_file': True}} <class 'kebechet.managers.version.version.VersionManager'> for 'thoth-station/kebechet', skipping
Traceback (most recent call last):
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.8/site-packages/ogr/abstract.py", line 107, in wrapper
    return function(*args, **kwargs)
  File "/home/user/.local/share/virtualenvs/user-5PivF_Od/lib/python3.8/site-packages/ogr/services/github/pull_request.py", line 163, in create
[...]

PR #1176, opened 2022-11-22
Use git add command instead of index.add from gitpython

harshad16:
Use git add command instead of index.add from gitpython
Signed-off-by: Harshad Reddy Nalla <hnalla@redhat.com>

fixes: #1174 

We are using git command method for add instead of index.add
as suggested here: https://github.com/gitpython-developers/GitPython/issues/292

## Description

With the method `repo.index.add(files)`
somehow that particular bit was include `.` files as well
which was causing in commits.

```
stderr: 'remote: error: object <xxx>: hasDot: contains '.'        
remote: fatal: fsck error in packed object        
error: remote unpack failed: index-pack abnormal exit

```

### DIFF
--- a/kebechet/managers/manager.py
+++ b/kebechet/managers/manager.py
@@ -177,9 +177,9 @@ pipenv version: {pipenv_version}
         cur_branch = self.repo.active_branch
         self.repo.git.checkout("HEAD", b=branch_name)
         try:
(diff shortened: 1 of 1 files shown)


harshad16:
Merging this to get a new release.

